PR #23, opened 2015-05-08
Add Releases page

### LINKED ISSUE #19: Add release notes page
Would be cool to have a nice looking release notes page similar to what atom.io has.

https://github.com/blog/1996-releases-metadata-for-github-pages has shipped which is cool for this.

https://atom.io/releases for inspiration

/cc @jlord

jlord:
This PR adds a Releases page to the site. 

Releases are fetched with `script/releases` and written to disk, `_data/releases.json`, and then used in `releases.md` and the template `releases.html`. 

It's linked to from the footer, which may or may not be noticeable enough? @kevinsawicki 

screen shot 2015-05-07 at 5 50 49 pm

Closes #19


kevinsawicki:
> It's linked to from the footer, which may or may not be noticeable enough?

I think that is a good start, it is how we have it on atom.io

_layouts/page.html
```diff
@@ -1,14 +1,7 @@
...
-  <header class="post-header">
-    <h1 class="post-title">{{ page.title }}</h1>
-  </header>
-
-  <article class="post-content">
-    {{ content }}
-  </article>
-
-</div>
+    <div class='page-template {{ page.title }}'>
```
kevinsawicki: Just curious, and super minor, but is there a reason this is indented?

script/releases
```diff
@@ -0,0 +1,17 @@
...
+
+var options = {
+  url: 'https://api.github.com/repos/atom/electron/releases',
+  json: true,
+  headers: { 'User-Agent': 'Electron' }
+}
+
+request(options, function (error, response, body) {
+  if (error) return console.log(error)
+  if (!error && response.statusCode == 200) {
```
kevinsawicki: I don't think you need to check for `!error` here, since the line before will return early any time `error` is truthy.

_layouts/page.html
jlord: No, I think it just ended up there as I went back in forth on trying to do it with/without templates. I've already got it moved back locally :+1: and will push up.

script/releases
```diff
@@ -0,0 +1,17 @@
...
+    fs.writeFileSync('_data/releases.json', JSON.stringify(body, null, ' '))
```
kevinsawicki: Super minor, but usually we use `2` as the third param to `JSON.stringify` so that we are always using a 2-space indent, a one space indent here makes `_data/releases.json` look a little strange.

script/releases
jlord: Oh, yeah :+1:

jlord:
http://electron.atom.io/releases.html :sparkle:

kevinsawicki:
> http://electron.atom.io/releases.html :sparkle:

Awesome, looks great :fireworks:

PR #26, opened 2015-05-14
Node.js -> io.js

lisposter:
Now, electron is using io.js :smile: .


jlord:
Fair point! :smile_cat: But as things are a bit in flux and as of yesterday [io.js is renaming to nodejs](https://twitter.com/rvagg/status/598605393636429825) and moving to the Node Foundation I think sticking with 'node.js' is the best bet right now.

lisposter:
:+1:  yes, I saw this after sent my pr. Anyway, that's awesome!

PR #27, opened 2015-05-16
Changed Spark as Particle 

muhasturk:
See http://blog.particle.io/2015/05/13/spark-is-now-particle/


kevinsawicki:
Thanks for this, I don't think the icon needs to be updated though since it looks like the Particle Dev app icon is the same as before.

I'm going to merge this and then go back to the old icon.

PR #32, opened 2015-06-02
Add Docs 'all' Page

jlord:
This adds an 'all' page with all of the latest version of docs on one page, accessible at [electron.atom.io/docs/all](electron.atom.io/docs/all).

It is linked to from the [electron.atom.io/docs](electron.atom.io/docs) (versions) page.  

screen shot 2015-06-02 at 1 35 23 pm


all.md
```diff
@@ -0,0 +1,19 @@
...
+category: ignore
+---
+
+# All of the Electron Documentation
+
+Below is the latest version of documentation. See all [available versions](/docs/).
+
+---
+
+{% for doc in site.docs %}
```
kevinsawicki: Oh nice, didn't know this was a possibility to making all in one pages. :+1:

jlord:
:mega: shoutout to @muan, Certified Jekyll Pro™, for the Liquid :droplet: help :raised_hands:

all.md
jlord: Yeah, it's nice for looping through your collections. The tricky bit was that if you have subdirectories within your collection there isn't a nice way to distinguish between them. 

Because `/latest` is a dupe of `/v0.27.0` there wasn't an easy way to filter them so that we just get one set (we'd get two of every document if we tried to filter on version since they have the same front matter). But they do have different file paths, and so that's what we use to get just one set of the latest to show. :sparkles:

PR #42, opened 2015-06-05
Fix Internal Link Paths

jlord:
Links inside of the docs that point to other docs (the relative links, rather than external) are broken because the relationship of the files is different on the website than it is in the `atom/electron` repo. 

This PR adds a script that parses all the relative links within the docs and fixes them so that they work relative to how the site is laid out. 

In other PR I'll work on resolving the links in the /all page. 

I'll also PR each update doc version in a separate PR.


kevinsawicki:
Awesome, nice work :+1:

jlord:
The absolute path generation isn't handling relative links in docs sub folders just yet.

PR #43, opened 2015-06-05
Update Internal Links v0.27.0

jlord:
This runs the fix in https://github.com/atom/electron.atom.io/pull/42 on version 0.27.0 of the docs, fixing the paths to internal links.


jlord:
Closing because I'm going to do absolute paths instead :sparkles:

PR #49, opened 2015-06-12
Fix doc script assumptions about OS path separator

atifaziz:
The script to fetch docs assumed that the OS path separator is always a forward slash (`/`) and while that works for most *nix-based systems, it was failing on Windows. This PR fixes that by using [`path.sep`](https://nodejs.org/api/path.html#path_path_sep) from Node wherever possible.


jlord:
Ah, yes, thank you! :sparkles:

PR #55, opened 2015-07-01
Remove assets/authors from releases data

kevinsawicki:
I noticed when running `script/releases` that `releases.json` would update often since the `download_count` on the `assets` entries would increment over time. This creates a diff even when there are no changes to the releases that should require a rebuild.

This pull request just drops the assets and author completely from the JSON file since they are unused.


jlord:
Yay!! Thanks :tada:

PR #69, opened 2015-08-11
Add Permalinks to Headers

jlord:
This PR adds the scripts and liquid templates that Jekyll needs to turn headers into _headers with permalinks_. [Info on how here](https://george-hawkins.github.io/basic-gfm-jekyll/redcarpet-extensions.html#hover-anchors).

I'm not :100: on having hidden link icons that you have to hover to see so if no one else hates them I've got them to the right of headers like so:

screen shot 2015-08-11 at 9 03 06 pm

Closes https://github.com/atom/electron.atom.io/issues/34

### DIFF
--- a/_includes/anchor_links.html
+++ b/_includes/anchor_links.html
@@ -1,0 +1,33 @@
+<script>
+  var anchorForId = function (id) {
+    var anchor = document.createElement("a");
+    anchor.className = "header-link";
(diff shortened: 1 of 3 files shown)


kevinsawicki:
:+1:

jlord:
:tada:

PR #82, opened 2015-09-09
Add Docs at v0.32.0

jlord:
Incoming docs updates! The latest docs are now here that have been standardized and go nicely with some new styles on the site:

screen shot 2015-09-09 at 9 33 59 am

screen shot 2015-09-09 at 9 34 11 am


jlord:
Oops, also want to note we have this fancy new notation implemented:

```text
### method(arg) _Platform_
```

Which turns into this on the website, noting when a method exists only on a certain platform:

screen shot 2015-09-10 at 12 25 55 pm

kevinsawicki:
These look really nice, great job :+1:

PR #86, opened 2015-10-05
Add Nylas N1 icon

kevinsawicki:
Just launched today, looks awesome, https://www.nylas.com/n1

/cc @bengotow for :+1:


bengotow:
Looks great! Fire away :+1:

PR #91, opened 2015-10-14
Added spreaker studio for desktop to developed apps

lele85:
@spreaker we built a cross platform audio streaming and recording app with electron. It's been a pleasure working with this awesome framework. :)


jlord:
Awesome! Thank you :sparkles: :speaker:

PR #99, opened 2015-10-23
Removed space after links

alajarvela:
The spaces that followed the links triggered an underline after each
image when hovering the images.

I removed the spaces but this messes up the indentation used there. This could also be fixed by adding a class to the div they're all inside and writing a specific CSS rule that removes text decoration for links inside that.

### DIFF
--- a/index.html
+++ b/index.html
@@ -86,104 +86,79 @@ layout: default
           <hr>
 
           <a href='https://atom.io' target='_blank' title='Atom'>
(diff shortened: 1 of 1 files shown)


jlord:
> This could also be fixed by adding a class to the div they're all inside and writing a specific CSS rule that removes text decoration for links inside that.

Ah, nice. This is a good fix for now, thanks! :+1:

PR #100, opened 2015-10-24
Add Light Table to "built on" and featured open source app lists.

kenny-evitt:
Add Light Table to apps lists.

I wasn't sure what the the first list was at first – I thought it might just be a list (of logos) of organizations using Electron. There doesn't seem to be any other apps in both lists so maybe Light Table should only be 'listed' in the first list?


kenny-evitt:
@cldwalker @rundis FYI

cldwalker:
@kenny-evitt Thanks for doing this. Happy to help promote Electron if they want us on the list :)

jlord:
Thanks!

PR #103, opened 2015-10-27
Scroll appears even when you do not need.

rafaelstz:
Before
image
image

After
image
image


thedaniel:
Thanks!

rafaelstz:
:+1: :smile: @thedaniel

PR #105, opened 2015-10-28
Add SteelSeries Engine 3 to app list

ToadKing:
We recently started using Electron in our Engine program with great results.


jlord:
Awesome, thanks!

PR #111, opened 2015-10-30
Exclude files not needed from the build process

benbalter:
This adds some excludes to Jekyll's config file to speed up the build process, especially `node_modules`, which doesn't appear to be used in the built site (Jekyll won't watch for changes in these folders or try to copy the files over to the site's output directory).

Testing locally, the build time went from about a minute to about 15 seconds.

/cc @jlord

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -27,3 +27,16 @@ collections:
     docs: {output: true, permalink: '/docs/:path/'}
 defaults:
     - {scope: {type: docs}, values: {layout: docs}}
+
+# exclude:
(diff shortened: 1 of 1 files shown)


jlord:
SO EXCITE! NEW LIFE HEREAFTER! 

:pizza: :tada: Thank you @benbalter!

cc @muan @kevinsawicki for FYI on how my builds went from 164s to 16s

maxbrunsfeld:
:clap: Thanks for the fix @benbalter

PR #112, opened 2015-11-12
Add Electron, Node, Chromium, V8 Versions to Home Page

jlord:
This adds a script and small section to the front page to display what the latest version of Electron is and what versions of Node, Chromium and V8 it's using :computer: 

Currently it'll just be manually run when we do a release (or maybe just `minor` releases) but :soon: I'll put this and the docs and releases data into a single thing that's run after a hook :fishing_pole_and_fish: 

screen shot 2015-11-12 at 9 14 15 pm

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,6 +11,9 @@ redcarpet:
 gems:
     - jekyll-redirect-from
 latest_version: v0.34.0
+current_node: 4.1.1
+current_chrome: 45.0.2454.85
+current_v8: 4.5.103.29
 available_versions:
     - v0.24.0
     - v0.25.0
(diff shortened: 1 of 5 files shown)


kevinsawicki:
Looks good :+1:

paulcbetts:
Cool Tricks Cool Tricks, here's how to get that info via electron-prebuilt:

https://github.com/electronjs/electron-rebuild/blob/master/src/main.js#L44

jlord:
@paulcbetts I don't get it :confused:

paulcbetts:
@jlord If you spawn `electron` with the `ATOM_SHELL_INTERNAL_RUN_AS_NODE` env var set, it'll act like `node`, so you can have it just do stuff and exit, like print `process.versions.modules` to output

jlord:
@paulcbetts ah, I see, I see. Good to know!

teobtcvn:
vâng...tuyệt vời

PR #123, opened 2015-12-06
added Laverna to list

bnvk:
Adding the open source note taking / organizing (Evernote-like) app [Laverna](https://laverna.cc) as they just added Electron support. It's not mentioned on their site yet, but is on [their Github releases](https://github.com/Laverna/laverna/releases/tag/0.7.0)


jlord:
:wave: @bnvk! Thanks!

PR #124, opened 2015-12-07
add Tweetman to app list

kkosuge:
Adding the Tweetman to Electron apps section.
I submitted it to Mac App Store.
https://tweetman.kksg.net/en


jlord:
:sparkles: Thanks!

PR #130, opened 2016-01-02
Adding Hozz to Built on Electron section

ppoffice:
Hozz is an open source `/etc/hosts` file manger for Windows, OS X and Linux. Hozz is built on Electron.
[Landing page](http://ppoffice.github.io/Hozz)


jlord:
:sparkles:

PR #144, opened 2016-02-07
Adding Turbo Download Manager to Built on Electron section

inbasic:
Turbo Download Manager is now available for Windows and Mac operation systems;
https://sourceforge.net/projects/turbo-download-manager/files/


jlord:
Thanks!

PR #147, opened 2016-02-11
Speak no longer live

muhasturk:
... wont accept signups.

See: http://blog.speak.io/epilogue/


jlord:
Aw, thanks @muhasturk :+1:

PR #160, opened 2016-02-22
Add Presets to Built with Electron section

Kaizhi:
[Presets.io](http://www.presets.io) is a desktop app built on electron to help users manage (install, review, uninstall) Adobe Lightroom presets.


jlord:
Super cool, thanks @Kaizhi!

PR #174, opened 2016-03-11
Parse Documentation for Footer Links

jlord:
All of the Electron documentation hosted in the repository on github.com is written in markdown and github.com handles any relative links to work correctly on the site.

We use the same docs here on this site so we need to parse for any links and make them work here, too. This pull request updates the parsing so that two forms of markdown links are acceptable. 

- Traditional: (in the body) `[i am a link](github.com)`
- Footer: (in the body) `[i am a link][i-am-link]` (in the footer) `[i-am-link]: https://github.com`

Specs are updated as well :pizza: 

cc @kevinsawicki :pear:

### DIFF
--- a/lib/doc-links.js
+++ b/lib/doc-links.js
@@ -8,7 +8,7 @@ var path = require('path')
 // to other pages within the docs (relative links,
 // rather than external).
 //
-// We're looking for these patterns:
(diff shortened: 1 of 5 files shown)


kevinsawicki:
:footprints: :+1:

PR #180, opened 2016-03-14
Add Egret Wing to "Build on Electron" section

zcxp:
Egret Wing is an IDE for HTML5 games, we have over 15000 developers. Now we have rewrite it with Electron.

Egret Wing:http://www.egret.com/products-wing
Game Engine: https://github.com/egret-labs/egret-core


jlord:
Thanks!

PR #188, opened 2016-03-18
issue1000 fixed, added Brave to the list of electron apps

viswanathamsantosh:
Fixing the issue < https://github.com/brave/browser-laptop/issues/1000 >


jlord:
Awesome, thanks!

PR #192, opened 2016-03-23
add notes about how to submit an app to the homepage list

zeke:
Hopefully this will help reduce the work required by the team to manage incoming submissions.

What do you think, @jlord?


jlord:
Awesome! Looks good 🐑

PR #193, opened 2016-03-23
add link to /docs/all

zeke:
http://electron.atom.io/docs/all is a thing! This adds a link to it in the docs footer:

screen shot 2016-03-23 at 3 27 20 pm

The icon layout needs a little love, but I'd rather :ship: and save that fix for another PR.


kevinsawicki:
:+1:

PR #200, opened 2016-03-31
Add Open Stage Control to app list

jean-emmanuel:
Open Stage Control is a libre user interface editor designed for controlling OSC enabled softwares on stages and in studios. 
Thanks !

http://jean-emmanuel.github.io/open-stage-control/

logo
capture d ecran_2016-04-01_01-03-21


jlord:
Thanks!

PR #218, opened 2016-04-15
Add Datazenit app to "Built on Electron" section.

lauris:
Datazenit is database admin tool. Our GUI is built on top of Electron to provide cross-platform desktop app.


zeke:
Thanks, @lauris!

PR #227, opened 2016-04-26
Adding Spectrum (https://www.devspectrum.com) to project list

frankwiles:
Wanted to add our new product that is getting released on Thursday to the list. Thanks for building and maintaining Electron, without it Spectrum would never have even attempted to be cross platform.


jlord:
Cool, thanks!!

PR #235, opened 2016-05-07
Update head data with Google meta tag

loranallensmith:
Adding site verification data for Google. :tada:


kevinsawicki:
Thanks for adding this 🔏 🚢

PR #234, opened 2016-05-05
Add Plain Email icon

apancik:

PR #233, opened 2016-05-05
Add GF-Trader icon

gliheng:
Hi, I added our application icon to the list. It's the first securities trade application developed using electron and web technologies in China.

PR #231, opened 2016-05-03
Add Etcher

craig-mulligan:
Add [Etcher](http://www.etcher.io/) to built on atom.io section. 

Etcher is a safe & easy way to burn images to SD cards and USB drives.

PR #157, opened 2016-02-21
Add poi to Built on Electron section

magicae:
:)

po is a powerful browser for playing webgame _KanColle_. Now poi has got 350 stars in GitHub, and 35000+ players are using poi.

Repo: https://github.com/poooi/poi

PR #237, opened 2016-05-10
Move apps to data file

jc00ke:
Easier to update the list now.

Also fixed the app descriptions, so you can read what the app is w/o having to click the link.


jc00ke:
If this is merged then #157, #231, #233 and #234 might want to rebase and update their PRs too.

mnquintana:
Would you mind making this a JSON or YAML file? CSV is quite unwieldy and difficult to maintain.

kevinsawicki:
This was done as part of the Electron 10 relaunch that just happened and was merged in #238 

There is now an `apps.yml` file that powers http://electron.atom.io/apps/

Thanks for opening this but I think we are all set now and I'm going to close this out. Apologies for the delayed response and thanks for contributing :shipit:

jc00ke:
Will you accept a PR that adds the description into the title attribute? IMO while it's nice one can see the name of the app, this new design still suffers from the same problem... I have to click on the link to find out what the app does.

kevinsawicki:
> Will you accept a PR that adds the description into the title attribute? 

Yup, that would be great, thanks.

PR #242, opened 2016-05-11
Add Plain Email to the list

apancik:
Just adding [Plain Email](http://www.plainemail.com) to the list! App available here: http://www.plainemail.com


kevinsawicki:
Thanks!

PR #245, opened 2016-05-11
Add Basecamp 3 to apps

javan:
Thank you for Electron! We'd love to be `featuredOnHomepage: true` sometime. ❤️


kevinsawicki:
Awesome, thanks 🚢

PR #251, opened 2016-05-11
Add app description to title attr in app grid

jc00ke:
Revamped #237

### DIFF
--- a/_data/apps.yml
+++ b/_data/apps.yml
@@ -41,7 +41,7 @@
   icon: "hive.png"
 -
   name: "Avocode"
-  description: ""
+  description: "One workspace open to all designers and developers"
   website: "http://avocode.com"
   repository: ""
   keywords: []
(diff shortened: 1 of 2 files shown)


kevinsawicki:
This looks good to me, @simurai @jlord any concerns?

simurai:
I like it too. 👍 

But maybe have name + description in the title? `{{ app.name }} - {{ app.description }}`.  So that apps with very long names are still shown somewhere.

screen shot 2016-05-12 at 3 27 49 pm

kevinsawicki:
> But maybe have name + description in the title? 

Yeah, good point, @jc00ke would you mind incorporating that?

jc00ke:
Sure, I actually had it that way originally but removed the app name. Will adjust shortly.

simurai:
Looks good! 🚢  Thanks @jc00ke

jc00ke:
Thank you!

PR #252, opened 2016-05-11
Add The Poker Timer 

karolk:
Our application is used by poker players to manage their tournaments. We are used by over 25k people worldwide. It used to be Adobe AIR app but we have migrated to Electron now.


kevinsawicki:
Do you have a `256x256px` version of your icon?

karolk:
I do, will send a commit later today.

karolk:
@kevinsawicki uploaded large poker timer icon.

PR #264, opened 2016-05-12
Update GitKraken app listing

chilblane:
Updated, larger icon, updated HTTPS url, and added fields.


lee-dohm:
Thanks for the update!

PR #272, opened 2016-05-14
Fix capitalization for OS X

stevekinney:
☺️

### DIFF
--- a/_docs/development/build-instructions-osx.md
+++ b/_docs/development/build-instructions-osx.md
@@ -1,7 +1,7 @@
 ---
 version: v1.0.1
 category: Development
-title: 'Build Instructions Osx'
+title: 'Build Instructions OS X'
 redirect_from:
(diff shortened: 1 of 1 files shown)


_docs/development/build-instructions-osx.md
```diff
@@ -1,7 +1,7 @@
 ---
 version: v1.0.1
 category: Development
-title: 'Build Instructions Osx'
+title: 'Build Instructions OS X'
```
kevinsawicki: This page is regenerated automatically so it needs to be fixed in here: https://github.com/electron/electron.atom.io/blob/2c099dd20c40765d7e1fde3c289f359224844a4f/lib/fetch-docs.js#L170

kevinsawicki:
This was fixed by #283 which now generates the titles from the headings instead of the filename.

PR #275, opened 2016-05-16
Updated to reflect the new electron 1.1.0 changes

amilajack:
Not sure what the v8 version is so I couldn't add that


kevinsawicki:
@amilajack this is updated automatically via the `script/versions` script which runs when we publish a new docs version.

I've done that in #279, thanks for bringing this up though.

PR #287, opened 2016-05-18
Drop the leading v from Electron version

kevinsawicki:
Noticed on the homepage that the Chrome, V8, and Node versions don't have a `v` prefix but the Electron version did.

This pull request removes that and so Electron releases will show with labels like `1.1.0` instead of `v1.1.0`.

/cc @jlord @simurai


simurai:
👍  Yeah, this should help with "quicker readability". 🐎

PR #292, opened 2016-05-20
Sort docs by title

kevinsawicki:
Previously docs were sorted by their filename instead of their title.

This pull request switches it to generate a new `sort_title` config value when building the docs that will be used as the sorting key when iterating over them.

This should lead to consistent case-insensitive sorting on pages like http://electron.atom.io/docs/ and http://electron.atom.io/docs/api/

/cc @jlord

### DIFF
--- a/_pages/guides.md
+++ b/_pages/guides.md
@@ -12,7 +12,8 @@ breadcrumb: Guides
 <h2 class="docs-heading pb-3 mb-3"><span class="mega-octicon octicon-book pr-3"></span>Guides</h2>
 
 <ul class="docs-list">
-{% for doc in site.docs %}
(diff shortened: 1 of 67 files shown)


_pages/guides.md
```diff
@@ -12,7 +12,8 @@ breadcrumb: Guides
 <h2 class="docs-heading pb-3 mb-3"><span class="mega-octicon octicon-book pr-3"></span>Guides</h2>
 
 <ul class="docs-list">
-{% for doc in site.docs %}
+{% assign docs = site.docs | sort: 'sort_title' %}
```
jlord: ✨

jlord:
Nice 👍

PR #293, opened 2016-05-20
Fix typo in _data/processes.yml

yuya-oc:
- BrowserWindow is not listed on http://electron.atom.io/docs/ .
- http://electron.atom.io/docs/api/browser-window/ does not show its category.


jlord:
Nice catch, thank you!

PR #296, opened 2016-05-24
Add timetable menubar

willyb321:
Based on electron and [electron menubar](https://github.com/maxogden/menubar)


_data/apps.yml
```diff
@@ -1035,3 +1035,26 @@
     - "research"
     - "spreadsheet"
   icon: "dataproofer.png"
+
```
kevinsawicki: Please remove these blank lines, thanks.

willyb321:
Ok, done. Thanks.

_data/apps.yml
```diff
@@ -1035,3 +1035,12 @@
...
   icon: "dataproofer.png"
+-
+  name: "Electron TimeTable"
+  description: "TimeTable in your menubar!"
+  website: "https://github.com/willyb321/electron-menubar-timetable"
+  repository: "https://github.com/willyb321/electron-menubar-timetable"
+  keywords:
+    - "Timetable"
+    - "menubar"
+  icon: "icon.png"
```
kevinsawicki: This needs to be the name of the icon you are adding in this pull request, in this case `TTTIcon.png`

_data/apps.yml
```diff
@@ -1035,3 +1035,12 @@
     - "research"
     - "spreadsheet"
   icon: "dataproofer.png"
+-
+  name: "Electron TimeTable"
```
kevinsawicki: Mind removing the `Electron` prefix from the name and just call it `TimeTable` instead?

kevinsawicki:
The current icon looks more like an avatar than an app icon, do you have a more application style icon you could use instead?

willyb321:
Alright, all that is done!

PR #301, opened 2016-05-25
Only create redirect URLs for older versions

jlord:
We create redirect urls for each page in the docs because we've changed what docs are on the site and where they are. This way old URLs that persist in Google searches will still take the user to the correct spot.

We were generating this URL based on all versions of Electron listed in the `_config` but the old urls only exist for versions prior to `1.0`. 

This PR updates the script so that we're only creating redirect URLs for versions prior to `1.0`.

You can see the result of this change in the new docs PR #302.

/cc @kevinsawicki


kevinsawicki:
Looks good 👍

PR #323, opened 2016-06-02
Revise app submission re: websites

jlord:
This PR makes an update to the app submission guidelines to specify pointing to a website or repository with a `README` . 

This is to insure we don't link directly to downloads or pages with just downloads in order for users to have a bit more information before they decide to install. 

This also points people to [GitHub Pages](pages.github.com) as an option for a free website 💻

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ Add your app to the list by editing [_data/apps.yml](/_data/apps.yml).
 * The required fields are `name`, `website`, and `icon`. Everything else is
   optional, but recommended.
 * Your `website` can be a repository with a `README`, but please provide a destination with information about your app.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -23,7 +23,8 @@ Add your app to the list by editing [_data/apps.yml](/_data/apps.yml).
 
 * The required fields are `name`, `website`, and `icon`. Everything else is
   optional, but recommended.
-* If your app doesn't have a website, you can specify a `repository` URL instead.
+* Your `website` can be a repository with a `README`, but please provide a destination with information about your app.
+  * You can make a free website with [GitHub Pages](pages.github.com)
```
kevinsawicki: I think this might need to be https://pages.github.com

PR #325, opened 2016-06-02
Add spec for 16 featured apps

kevinsawicki:
Sometimes pull requests add `featuredOnHomepage: true` in their pull requests and since we aren't adding new ones at this time, this pull request adds a spec verifying exactly 16 are featured so we know the right amount appear on http://electron.atom.io/

/cc @jlord


jlord:
Very nice! ✨

PR #326, opened 2016-06-03
href missing fixed

Siddharth11:
The href attribute was missing for last 3 added apps - Caprine, Anatine and Shiba.


jlord:
Oh, sorry, I should have started at bottom of list but I got to https://github.com/electron/electron.atom.io/commit/b344c90c5ef89ea1609584811bfee8d3b2353ca9 first which made these changes. Thanks so much though!

PR #328, opened 2016-06-03
Add Cypress, automated test runner

jennifer-mann:
Suggest adding Cypress to apps page.

The app is still in beta and available for users by closed invite system. Public release to come this year, but many users are already using it in Mac & Linux.


jlord:
Cool, thanks!

PR #332, opened 2016-06-06
Add Subordination app

sunabozu:
It's an Electron application for translating and editing subtitles. Would be nice to see it on your list.


jlord:
Thanks!

PR #343, opened 2016-06-13
Escape app descriptions

### LINKED ISSUE #342: alt text is cut off if app description contains an apostrophe
screen shot 2016-06-13 at 10 47 21 am

`<a href='https://www.getflow.com' target='_blank' title='Flow - Task Management Software That Doesn't Feel Like Work' class='featured-app'>`

vi @as-cii

jlord:
This adds `| escape` to the places where we are using `app.description` which may contain characters that break how the text is parsed. 

screen shot 2016-06-13 at 10 54 43 am

cc @zeke @as-cii

Closes #342


jlord:
I thought this was using an emdash like in the other section, but then realized it wasn't when I looked again at my screenshot, so just set it back to dash 🏃

PR #347, opened 2016-06-15
fix spell error

banyudu:


jlord:
Thanks again, same as in #348 though, the changes need to me made in [electron/electron](https://github.com/electron/electron) 👍

PR #348, opened 2016-06-15
fix spell error

banyudu:


jlord:
Thanks for finding and fixing this! The documentation is actually [pulled directly from the electron/electron repository](https://github.com/electron/electron.atom.io/blob/gh-pages/CONTRIBUTING.md#documentation) and so documentation changes need to be made there.

[electron/electron](https://github.com/electron/electron)

PR #351, opened 2016-06-18
Add devRant app

Meadowcottage:
Pull request to add unofficial devRant desktop app to app list.

Official Site: https://www.devrant.io/


_data/apps.yml
```diff
@@ -1298,3 +1298,13 @@
     - "Unity"
     - "Scratch"
   icon: "gausssense.png"
+-
+  name: "devRant.io"
```
jlord: I wonder if we should include "(Unofficial)" here and in the description also?

Meadowcottage: Updated the description to state that it is an unofficial app.

jlord: If you aren't affiliated with `devRant.io` I think you should add it to the `name` on line `1302` as well. Ready to merge after that 👍

_data/apps.yml
```diff
@@ -1298,3 +1298,13 @@
...
+  description: "Unofficial devRant app that let's you connect with fellow developers over fun rants about tech"
```
50Wliu: It should be `lets` and not `let's`

Meadowcottage: Typo fixed.

_data/apps.yml
Meadowcottage: Added `(Unofficial)` to name.

jlord:
Thanks!

PR #355, opened 2016-06-22
Update site to v1.2.4

jlord:
New version of Electron! :pizza:


zeke:
I've been missing those 🍕 s, @jlord. Glad to see you still party!

PR #359, opened 2016-06-29
Add link to doc translations

jlord:
I don't think we surface anywhere on the site the fact that there are some doc translations. Those translations aren't as up to date as the ones on the site but I thought it was maybe still helpful to expose? cc @zeke 

screen shot 2016-06-29 at 1 52 31 pm


jlord:
> We could also potentially mention here that help is always needed in the translation department.

Good idea, I think adding a `README.md` to that directory in `electron/electron` would be a good place for that.

PR #367, opened 2016-07-07
Fix a hash of a link href in the remote.md

vsemozhetbyt:
web-contents.md#webcontentsexecutejavascriptcode-usergesture ->
web-contents.md#webcontentsexecutejavascriptcode-usergesture-callback

### DIFF
--- a/_docs/api/remote.md
+++ b/_docs/api/remote.md
@@ -63,7 +63,7 @@ win.loadURL('https://github.com');
 ```
 
 **Note:** for the reverse (access the renderer process from the main process),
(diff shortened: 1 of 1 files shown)


jlord:
Thanks @vsemozhetbyt! I see you made the change in `electron/electron` which is where these docs are pulled from; so that's all that needs to be done. That change will get incorporated here in the next release.

vsemozhetbyt:
Oh, I see. Sorry.

jlord:
No worries! Thanks for improving the docs!

PR #373, opened 2016-07-08
Add spec for missing process categorization

kevinsawicki:
All API docs should be labeled with the process they should be used in, either `Main`, `Renderer`, or `Both`.

There were two pages missing them currently and this pull request adds entries to `processes.yml` for them.

Also adds a spec that verifies that all API pages have associated process categories.


_docs/api/content-tracing.md
```diff
@@ -36,7 +36,7 @@ redirect_from:
     - /docs/latest/api/content-tracing/
 source_url: 'https://github.com/electron/electron/blob/master/docs/api/content-tracing.md'
 excerpt: "Collect tracing data from Chromium&apos;s content module for finding performance
-bottlenecks and slow operations."
+    bottlenecks and slow operations."
```
kevinsawicki: This was required so that `frontmatter.loadFront(apiDoc)` would load correctly, it was failing with a parse error if this was outdented fully.

PR #376, opened 2016-07-08
Remove permalink for FAQ

jlord:
When the next version of docs ships the FAQ will live at this location so we don't need to use the permalink anymore ✨ 

cc @zeke I think this is all we need to change to prep for that update?


zeke:
Sweet. So should we ship this now, then remove it later?

jlord:
> then remove it later?

Not sure what the 'it' is here 😄 This PR does remove the permalink.

zeke:
🙈 oops I misread. So we should wait until 1.2.7 to merge this?

jlord:
We can merge it now since these scripts are running on the site.

PR #384, opened 2016-07-15
Adding Marp app

yhatt:
[Marp](https://yhatt.github.io/marp/) is the simplest presentation writer with Markdown. Could we add it?


zeke:
I saw Marp on awesome-electron the other day. Pretty nifty.

It's the first Electron app I've used that has one of those helpful "drag it to applications" installation windows on macOS, so bonus points for that.

screen shot 2016-07-12 at 8 53 24 pm

yhatt:
@zeke Thank you for your compliment!

PR #386, opened 2016-07-18
Add Ghetto Skype to app list

stanfieldr:


lee-dohm:
Thank you very much for your submission for inclusion in the Electron app list. We're uncomfortable with including your application in the list because of the loaded term "ghetto" especially combined with the use of someone else's product name in your project's name. Because of these reasons, we've decided to not include your application in our list.

Thanks very much for your submission and your use of Electron!

PR #398, opened 2016-07-27
Added Electorrent to apps.yml

Tympanix:
Added my new electron app to apps.yml and uploaded an application icon compressed [(tinyPNG)](https://tinypng.com/) to 12KB and 256x156 in size.


jlord:
Thanks @Tympanix!

PR #401, opened 2016-07-27
apply OS label style to H4 in addition to H3

zeke:
These styles regressed when we updated the styleguide and docs to use H4-level headings for things like Events and Instance Methods.

screen shot 2016-07-27 at 2 07 03 pm

👀  @kevinsawicki


kevinsawicki:
👍 Looks good, thanks for fixing this 👍

PR #410, opened 2016-08-04
Blog post on using node as library

zcbenz:
The second post of Electron Internals. Feel free to improve the text directly.

PR #418, opened 2016-08-08
Add color to links within <code>

jlord:
When previewing #410 I noticed that links that were wrapping `<code>` tags didn't get any new style so it was hard to tell they were links at all. This gives them the same colors as text links and you can now 👀 them. 

| Before | After |
|---|---|
| img | img2 |

---

screen shot 2016-08-08 at 10 53 09 am


jlord:
I can't make a markdown table today.

PR #420, opened 2016-08-08
Style platform labels in lists

### LINKED ISSUE #419: Platform labels not styled for function parameters
The `macOS` label is not styled on http://electron.atom.io/docs/api/auto-updater/#autoupdatersetfeedurlurl-requestheaders

screen shot 2016-08-08 at 12 54 16 pm

### LINKED ISSUE #421: Remove upper-case for platforms labels
With the change from `OS X` to `macOS`, I think the [all upper-case](https://github.com/electron/electron.atom.io/blob/gh-pages/_sass/_docs.scss#L92) should be removed. The `MACOS` feels harder to recognize. When removed it would look like this:

screen shot 2016-08-10 at 1 11 56 pm

jlord:
Platform labels were not getting styled anywhere other than in method/event headers. There are platform labels, however, in the lists of arguments for methods and those need to be styled.

There _shouldn't_ be anything else we're using `em` for in a list in the docs (I think??) so what I have now may be fine: `.docs li em` but I'd like @simurai's thoughts. Should it be more specific? 

🔗 [to live doc](http://localhost:4000/docs/api/auto-updater/#autoupdatersetfeedurlurl-requestheaders)

before

after

Fixes #419 
Fixes #421

### DIFF
--- a/_sass/_docs.scss
+++ b/_sass/_docs.scss
@@ -90,7 +90,6 @@
     vertical-align: middle;
     font-size: .66em;
     font-weight: 300;
-    text-transform: uppercase;
     font-style: normal;
     color: $main-bg-color;
     border-radius: $border-radius;
--- a/_sass/_docs.scss
+++ b/_sass/_docs.scss
@@ -84,7 +84,8 @@
 .docs h3 code > em,
 .docs h3 > em,
 .docs h4 code > em,
-.docs h4 > em {
+.docs h4 > em,
+.docs li em {
     padding: .1em .4em;
     vertical-align: middle;
     font-size: .66em;
--- a/_sass/_docs.scss
+++ b/_sass/_docs.scss
@@ -85,7 +85,7 @@
 .docs h3 > em,
 .docs h4 code > em,
 .docs h4 > em,
-.docs li em {
+.docs .platform-label {
     padding: .1em .4em;
     vertical-align: middle;
     font-size: .66em;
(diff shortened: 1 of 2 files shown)


_sass/_docs.scss
```diff
@@ -84,7 +84,8 @@
 .docs h3 code > em,
 .docs h3 > em,
 .docs h4 code > em,
-.docs h4 > em {
+.docs h4 > em,
+.docs li em {
```
kevinsawicki: Looks like there are some other non-platform label elements matching this selector.

I ran 
```js
document.querySelectorAll('.docs li em')
```

on http://electron.atom.io/docs/all/

jlord: Oh yeah, nice, I totally forgot about that page. Hrmm.

simurai:
If it still should stay Markdown only (and not a span with a class), how about use a convention of **bold** and _italic_ to identify the platform? Like `**_macOS_**`

- `requestHeaders` Object **_macOS_** - HTTP request headers.

Selector would be `.docs strong > em`. Then it could be used in more places, not just header and lists, but also normal paragraphs or quotes. The combination bold + italic doesn't seem to be used anywhere yet:

```js
document.querySelectorAll('.docs strong > em')
```

jlord:
@simurai Thank you! We just talked it over and decided rather than change the syntax we're using for platform labels we would just use JS on the client-side to find those labels (italic elements with `textContent` matching the platform name) and add a class to the `em` element. 

I also went ahead and took out the `text-transform` from your #421 ✨

PR #424, opened 2016-08-11
Add Ionic Creator & TwitchAlerts to apps list.

Meadowcottage:
Simply adding the apps to the app list along with their app icons.


jlord:
Thanks!

amitmerchant1990:
@jlord I think we need to add `website` field over here as right now it is redirecting to http://electron.atom.io/apps/ when clicking on both of these icons.

PR #429, opened 2016-08-16
move application icon of seapig to right path

yasumichi:
sorry, I deployed application icon wrong path. Please, move to right path.


jlord:
Whoops, good catch, thanks!

yasumichi:
Thanks you.

PR #432, opened 2016-08-17
Added website attribute to Ionic Creator & TwitchAlerts

Meadowcottage:
A quick fix suggested by @amitmerchant1990 to add the website attribute as at the moment the link doesn't go anywhere apart from the current apps page.


jlord:
Thanks 👍 

I'll also update the page itself to use either website or repository.

PR #433, opened 2016-08-17
On apps page use repository URL if there is no website

jlord:
As @amitmerchant1990 pointed out https://github.com/electron/electron.atom.io/pull/424#issuecomment-240324301, not all apps provided a `website` so the linked image just refreshed the page.

This PR updates the template to use the `repository` URL if no `website` is provided. 

Thanks @muan for this ProTip™ ⭐ `{{ app.website | default: app.repository }}`


amitmerchant1990:
Thanks for fixing this @jlord :)

PR #435, opened 2016-08-18
Fixed "in the future" for 32bit

WillEccles:
Added "the" to "in the future."

### DIFF
--- a/_docs/development/build-instructions-osx.md
+++ b/_docs/development/build-instructions-osx.md
@@ -90,7 +90,7 @@ After building is done, you can find `Electron.app` under `out/D`.
 ## 32bit Support
 
 Electron can only be built for a 64bit target on macOS and there is no plan to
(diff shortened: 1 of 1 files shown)


jlord:
Thanks for catching this! The docs on this site, however, are copied from the source at `electron/electron` and so [changes need to be made upstream there](https://github.com/electron/electron) ✏️ 

[Here's the bit on that](https://github.com/electron/electron.atom.io/blob/gh-pages/CONTRIBUTING.md#documentation) in the `CONTRIBUTING.md`.

jlord:
Oh, just saw you already made the pull request there: https://github.com/electron/electron/pull/6895#issuecomment-240769483

Thanks! When the next release 🚢 the doc will be updated here ✨

WillEccles:
@jlord oh, sorry, didn't realize that's how that works! should have read about it. thanks!

PR #437, opened 2016-08-19
Show process in API Reference table

jlord:
This adds the process/s name in the table of Electron APIs at [electron.atom.io/docs/api](http://electron.atom.io/docs/api).

I added a class to fit it all in there but don't know if it should be done in a different way so pinging @simurai for thoughts ✨ 

| Before | After |
| --- | --- | 
|screen shot 2016-08-19 at 2 07 24 pm|screen shot 2016-08-18 at 1 33 20 pm|

There is an After-After also... The "Process" and "Processes" text looks repetitive so I took it out but that meant needing to add table headers. What's your preference @simurai?

| After-After |
| --- | 
| screen shot 2016-08-19 at 2 13 52 pm|

### DIFF
--- a/_sass/_basecoat-tables.scss
+++ b/_sass/_basecoat-tables.scss
@@ -22,3 +22,10 @@
     white-space: nowrap;
   }
 }
+
+.table-with-spacious-second-column td:nth-child(2) {
+  @include breakpoint(md) {
+    padding-right: $spacer5;
+    white-space: nowrap;
+  }
+}
--- a/_sass/_basecoat-tables.scss
+++ b/_sass/_basecoat-tables.scss
@@ -16,13 +16,7 @@
   font-size: 14px;
 }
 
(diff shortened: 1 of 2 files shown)


simurai:
I like the "After-After" with the process/es removed.

_sass/_basecoat-tables.scss
```diff
@@ -22,3 +22,10 @@
     white-space: nowrap;
   }
 }
+
+.table-with-spacious-second-column td:nth-child(2) {
```
simurai: Not a big deal, but since the styles are the same, you could chain the selector, like:

```scss
.table-with-spacious-first-column td:first-child,
.table-with-spacious-second-column td:nth-child(2) {
  // ...
}
```

jlord: Oh yeah lol 😊

PR #456, opened 2016-09-01
New electron Application

scgm11:
Softphone WebRTC


jlord:
Hiya! I went through the motions too fast here and merged this, then found a couple of issues so I reverted the change on `gh-pages`

- The website `https://www.integraccs.com` does not appear to be `https` and should be updated.
- The `repository` field is for a URL to the location of the code for the app, if one exists, which is not what was provided in this branch. If there is no repository this field can be removed.

Could you make these updates and resubmit? Thanks! ✨

PR #458, opened 2016-09-03
Update apps.yml

scgm11:
fixed issues


zeke:
@scgm11 I think this PR can be closed, as it looks like https://github.com/electron/electron.atom.io/pull/456 already landed.

scgm11:
it was reverted

zeke:
Oh, I see. In that case, are you able to make the necessary updates and resubmit, like @jlord asked in https://github.com/electron/electron.atom.io/pull/456#issuecomment-244431419?

scgm11:
done this should be mergeable!

scgm11:
not yet, some minutes I correct all

zeke:
@scgm11 it looks like some old content (Wagon app) crept back into the PR. I opened a new one here for you: https://github.com/electron/electron.atom.io/pull/466

PR #464, opened 2016-09-06
Add roundup blog post helper

jlord:
This adds a little helper for generating markdown table rows based on apps in `_data/apps.yml` so that there is much less copypasta needed to do a roundup post. Automation! 💥 

This will mostly be used by myself and/or Electron team when working on these types of posts ✏️ 

screen shot 2016-09-06 at 4 07 09 pm

[Example roundup post](http://electron.atom.io/blog/2016/09/06/august-2016-roundup)

cc @zeke @kevinsawicki

### DIFF
--- a/lib/roundup-helper.js
+++ b/lib/roundup-helper.js
@@ -1,0 +1,21 @@
+var fs = require('fs')
+var yaml = require('yamljs')
+
+// Usage: node lib/roundup-helper.js <RoundupMonth> <PostMonth> <Year>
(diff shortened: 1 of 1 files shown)


zeke:
> Automation! 💥

Hear, hear.

Defaulting to last month if no args are given would be cool.

jlord:
> Defaulting to last month if no args are given would be cool.

Yeah, may tweaks this some more as I use it more.

PR #466, opened 2016-09-07
add uphone app

zeke:
A followup to #456 and #458


jlord:
🎉 Thanks! 

cc @scgm11

PR #478, opened 2016-09-19
Update links

torn4dom4n:
Fix `Page not found` and replace `http` with `https`


jlord:
Thanks @torn4dom4n! I made one small change, removing the year from the license file. When I asked our legal team if we needed to update the year they said we didn't need to worry about it at all 🎉

PR #489, opened 2016-09-25
Add Forestpin Analytics to app list

vpj:
I want to add our [Forestpin Analytics app](http://forestpin.com/analytics)


Haacked:
Looks good to me. Thanks! :sparkles:

PR #517, opened 2016-10-06
remove 1.0 banner from homepage

zeke:
Electron hit 1.0 six months ago. This PR removes the banner from the homepage.

screen shot 2016-10-06 at 10 21 24 am

@kevinsawicki @Haacked


Haacked:
Looks good to me. If the others have no objection. :smile:

_layouts/default.html
```diff
@@ -5,10 +5,7 @@
 
   <body>
     {% include header.html %}
-    {% if page.url == '/' %}{% include banner-1-0.html %}{% endif %}
```
simurai: 👍  on removing it.

Might still be worth it to rename it to `banner.html`, remove the copy and just comment out here. So that at some point in the future, when there is a need for a banner, it can quickly be added again.

PR #522, opened 2016-10-08
Added sencha electron products

adwankar:


Haacked:
Looks like the formatting of this is incorrect. Please fix up the formatting by looking at the other examples.

PR #524, opened 2016-10-09
Add Elite Journal

willyb321:
Hi, I'd like to add my application, Elite Journal to the apps list!  
https://github.com/willyb321/elite-journal  
Thanks!


Haacked:
Thanks for the submission.

PR #525, opened 2016-10-09
Removed TwitchAlerts from App List

Meadowcottage:
I have been asked by the people at StreamLabs (Previously known as TwitchAlerts) to remove my app from the app list so they can submit their own official one.


Haacked:
Cool, thanks!

PR #526, opened 2016-10-09
Remove two defunct apps from app list

charlesj:
- Pixate is shutting down and is no longer available
- antSword points to an empty git repository


Haacked:
Thanks for the cleanup work! ✨ 👍

PR #528, opened 2016-10-10
Update apps.yml

adwankar:
added sencha products

### DIFF
--- a/_data/apps.yml
+++ b/_data/apps.yml
@@ -2066,7 +2066,7 @@
     - "debugger"
   icon: "sencha_inspector.png"
 -
-    name: "Sencha Themer"
+  name: "Sencha Themer"
   description: "Theming tool to rapidly style Ext JS applications by creating custom themes using graphical tools – without writing code."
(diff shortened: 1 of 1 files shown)


_data/apps.yml
```diff
@@ -2055,3 +2055,34 @@
...
+  description: "Debugging tool for troubleshooting and improving performance of Ext JS and Sencha Touch applications."
+  website: "https://www.sencha.com/products/inspector/"
+  keywords:
+    - "web development"
+    - "programming"
+    - "productivity"
+    - "debugger"
+  icon: "sencha_inspector.png"
+-
+    name: "Sencha Themer"
```
Haacked: Yaml is whitespace sensitive. So this `name` line should not be indented.

adwankar: updated name line

zeke:
Thanks, @adwankar. Every app needs an icon to be featured on the site. Please add icons to the `images/apps` directory and reference each filename from `apps.yml`

PR #531, opened 2016-10-10
update atom/electron URLs to electron/electron

zeke:
cc @kevinsawicki @Haacked


Haacked:
Nice catch and pretty straightforward. 👍

PR #536, opened 2016-10-18
Add --shots

binjospookie:
Hello, I developed this app for better feedback between users.
It's simple and free.


kevinsawicki:
Merged in commit 3c98e05

PR #552, opened 2016-11-01
feature four more apps on homepage

zeke:
This PR adds four more featured apps to the homepage:

- https://getkap.co @matheuss @CodeTheory
- https://zeit.co/app @rauchg @leo
- https://beakerbrowser.com @pfrazee
- http://www.svgs.us @stonebranch

### DIFF
--- a/_data/apps.yml
+++ b/_data/apps.yml
@@ -490,7 +490,6 @@
   keywords: []
   license: ""
   icon: "googleplaymusicdesktopplayer.png"
-  featuredOnHomepage: false
 -
   name: "World History AP"
   description: "Help students study for the rigorous AP exam"
@@ -499,7 +498,6 @@
   keywords: []
   license: ""
   icon: "whap.png"
-  featuredOnHomepage: false
 -
   name: "Hain"
   description: "An 'alt+space' launcher for Windows"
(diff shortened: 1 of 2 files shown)


leo:
😍😘

pfrazee:
> 😍😘

Seconded!

gschier:
How is it decided which apps get featured?

zeke:
> How is it decided which apps get featured?

Our current approach is to try to choose apps that stand out as being high quality and well-maintained. We will continue to feature some apps on the website based on this criteria, but we are also working to provide more objective, data-driven reports. The new http://electron.atom.io/userland/ pages do just that: They surface interesting apps, packages, and users based on criteria like fork counts, contributor counts, download counts, number of dependent packages, etc.

gschier:
Thanks for the answer @zeke. I take it that means a close-source app will have much less chance of being included?

zeke:
> a close-source app will have much less chance of being included?

I wouldn't say so. We have not discriminated on that basis. Slack is a great example of that.

Do you have some app(s) in mind?

gschier:
I added [Insomnia](https://insomnia.rest/) (I'm the dev) to the apps page a while ago, and have seen other apps that were added after get featured, so I was curious. Obviously I would love for it to be featured, but I also respect whatever selection processes are in place. Thanks again.

alexk111:
@zeke Could you consider featuring [Freeter](https://freeter.io) on the website - developers love it.

PR #554, opened 2016-11-02
Replace Open Graph image

simurai:
This replaces the "1.0" graphic with the generic Electron logo.

See the [image diff](https://github.com/electron/electron.atom.io/pull/554/files).

That image will show up when people link to this site from Twitter, Facebook, Slack etc. And the 1.0 feels a bit outdated by now.


zeke:
🙇

PR #558, opened 2016-11-03
Encode pipes so they don't become tables

### LINKED ISSUE #556: Documentation: sometimes a long space appears between union type names
For instance, the following md snippet code:
```
The `net` module has the following methods:

### `net.request(options)`

* `options` (Object | String) - The `ClientRequest` constructor options.
```
may be rendered as follows:
image

Related link: https://github.com/electron/electron/pull/7862

zeke:
I pinged the GitHub Pages team about this one. @parkr points out that there are some deficiencies in the [kramdown](https://github.com/gettalong) markdown parser used by Jekyll:

> It uses regular expressions to match things. So if you have `|` then 10 lines later a `|`, the content in between may be considered a table.

It would be nice to land a fix for this in kramdown itself, but here is an interim fix. The line-by-line technique is used to match a specific pattern and avoid encoding pipes elsewhere in our markdown files.

cc @gettalong, maintainer of kramdown
cc @aliib, who reported the issue
cc @kevinsawicki, pull-request-reviewer extraordinaire

Fixes #556

### DIFF
--- a/lib/doc-links.js
+++ b/lib/doc-links.js
@@ -127,7 +127,7 @@ module.exports = function fixLinks (dir, version, callback) {
     // https://github.com/electron/electron.atom.io/issues/556
     content = content
       .split('\n')
-      .map(line => line.match(/^\* `/) ? line.replace(/ \| /g, ' &#124; ') : line)
+      .map(line => line.match(/^\s*\* `/) ? line.replace(/ \| /g, ' &#124; ') : line)
       .join('\n')
 
     writeLinkstoFile(file, content)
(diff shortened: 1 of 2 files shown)


lib/doc-links.js
```diff
@@ -122,6 +122,14 @@ module.exports = function fixLinks (dir, version, callback) {
     oldLinks.forEach(function (oldLink, i) {
       content = content.replace(oldLink, newLinks[i])
     })
+
+    // Work around kramdown's misinterpretation of pipe as a table header
+    // https://github.com/electron/electron.atom.io/issues/556
+    content = content
+      .split('\n')
+      .map(line => line.match(/^\* `/) ? line.replace(/ \| /g, ' &#124; ') : line)
```
kevinsawicki: Adding a `\s*` between the `^` and `\*` will replace pipes in sublists as well, not sure if we have any now but it seems possible that we could in the near future.

kevinsawicki:
:cool: Encoding these sounds good to me 👍

lib/doc-links.js
zeke: 👍

PR #562, opened 2016-11-06
support redirects for renamed docs

zeke:
This PR exists in anticipation of a rename in electron/electron:

`web-view-tag.md` 👉 `webview-tag.md`


zeke:
This will need to be revamped, given all the change that happened in #630.

zeke:
Replaced by #641

PR #569, opened 2016-11-12
Remove Whatsie

mimi89999:
The website says: "Whatsie is not available anymore.
Please use the official client."


zeke:
Thanks @mimi89999. Would you mind removing the icon file too?

mimi89999:
@zeke I will only be able to update that PR in 24h.

PR #587, opened 2016-11-23
Add missing titles and processes

kevinsawicki:
These were accidentally removed in #586 since some of the classes were switched to using `##` instead of `#` as their heading and had a `Class:` prefix added.

This updates the title parser to look for the first `<h2>` as well and drop the `Class:` prefix.


zeke:
Thanks!

PR #590, opened 2016-11-23
Put app image for GrabCAD Print into correct location

hmak3d:
Amends PR https://github.com/electron/electron.atom.io/pull/588 by putting icon in correct directory


kevinsawicki:
Thanks 👍 🚢

zeke:
🙊 good catch! Sorry I missed it.

PR #595, opened 2016-11-26
wheredat logo

odedhb:
to include a new app in _data/apps.yml


MarshallOfSound:
This should be a single PR, try adding this image as a second commit to your PR in #596

PR #628, opened 2016-12-15
replace mojibar with insomnia

zeke:
mojibar is not actively maintained, so let's remove it to make space for something new.

https://insomnia.rest/


gschier:
😸

PR #630, opened 2016-12-17
Build Cleanup

zeke:
Our docs build process was a bit messy. This PR cleans it all up.

- [x] Introduce a new [Doc Class](https://github.com/electron/electron.atom.io/blob/b127a8440753281a54393eee2e2a7aa67eb05597/lib/doc.js) for easier-to-read code.
- [x] Use [electron-docs](https://github.com/electron/electron-docs) to fetch markdown files. This is an existing (tested) module that the docs linter uses to collect the docs.
- [x] Remove version info from Jekyll's `config.yml` in favor of `_data`, as it's not really config
- [x] When building docs, always use the [highest version number](https://github.com/electron/electron.atom.io/blob/57ab9999149e3fa04736cc5a08c956f4cbb5185c/test.js#L109-L113), which is not necessarily the most recent release. This will prevent automated deploys from publishing docs for backported releases of Electron. HT @kevinsawicki
- [x] Simplify [rewriting of relative links](https://github.com/electron/electron.atom.io/blob/57ab9999149e3fa04736cc5a08c956f4cbb5185c/lib/doc.js#L18-L25) by doing it against compiled HTML instead of constructing error-prone markdown regex patterns
- [x] Update dev instructions in readme
- [x] Remove unused dependencies, and consolidate them all into `devDependencies` because this app has no "runtime" dependencies.
- [x] Update test suite to use `mocha` instead of `tape`, and test against the real data instead of fixtures.
  - [x] test apps
  - [x] test versions
  - [x] test releases
  - [x] test docs
  - [x] test userland


zeke:
@electron/maintainers this is ready for review.

kevinsawicki:
> +11,855 −10,245

This is going to make it really hard to review, could this pull request omit the checked-in documentation changes that the other changes in this pull request generate?

zeke:
> +372 −1,256 

Done!

zeke:
@kevinsawicki CI is failing because tests are evaluating stuff in the `/_docs` directory (which I removed to make the review process easier). Once you've reviewed, I can push up the built changes to get CI passing again.

Gruntfile.js
```diff
@@ -1,35 +0,0 @@
-var cheerio = require('cheerio')
-
-module.exports = function (grunt) {
-  grunt.loadNpmTasks('grunt-link-checker');
-
-  grunt.initConfig({
-    pkg: grunt.file.readJSON('package.json'),
-
-    linkChecker: {
```
kevinsawicki: I'd like to keep this file (and its `package.json` dev dependencies), I still use this to check that there are no broken links when running locally and tweaking how links are parsed.

_data/apps.json
```diff
@@ -648,6 +648,20 @@
     "icon": "explorer-icon.png"
   },
   {
+    "slug": "fangyuanjian",
```
kevinsawicki: Just curious, why is this app showing up in this diff?

Gruntfile.js
zeke: Will do.

_data/apps.json
zeke: Because I ran `npm run build-apps` at some point.

Gruntfile.js
zeke: Okay, I brought `Gruntfile.js` back, along with its deps. Can you describe how you use it, so I can add a note to the README?

zeke: All right, the link checker is back in place. It found two errors, one of which is [already fixed](https://github.com/electron/electron/commit/704433f3d620d5dcb49ad813848816615288372e) and another for which I have [opened a new PR](https://github.com/electron/electron/pull/8240).

That check is really handy. I would love to get it baked into the automated test suite somehow eventually.

zeke:
Shipped! Sorry for springing this big refactor on you @kevinsawicki and thanks for the review.

PR #639, opened 2016-12-21
fix link tests to account for electron docs changes

zeke:
https://github.com/electron/electron/pull/7878/files rewrites the URL of the `Main Process` and `Renderer Process` links in each API doc.  Once that lands in a release, a few tests will start failing here.

This PR exists in anticipation of that future...


zeke:
This fixed the test suite, and the bot was able to publish the new 1.4.14 docs.

PR #648, opened 2017-01-04
Removing London from meetups list

tommorris:
The London Meetups group seems to have disappeared. Probably ought to be removed from the website pending confirmation it exists.


zeke:
Thanks!

PR #652, opened 2017-01-11
Filterable App List

zeke:
This borrows the filtering feature from [userland](http://electron.atom.io/userland/most_downloaded_packages) and brings it to the app page.

The appearance of the app list itself is unchanged, for now. Though the description and keywords are not displayed, they are included in the page and used when filtering.

It also syncs the filter input with the URL query string, so you can bookmark or share a URL: `/apps/?q=music`

cc @electron/maintainers 

Eventually I want to display the descriptions and other metadata, but this is start. Iteration ftw!

screen shot 2017-01-11 at 10 16 02 am

screen shot 2017-01-11 at 10 16 18 am


zeke:
Pushed this to my fork: https://zeke.github.io/electron.atom.io/apps/?q=music

cc @simurai

simurai:
Sooo useful. 💯 

> Eventually I want to display the descriptions and other metadata

Are you thinking of showing that on hover or all the time? I kinda like the "wall of icons", but having to hover to read the descriptions needs a lot more effort. Maybe description could be shown only when filtering?

Also saw that the list is now alphabetically ordered. More 11111aaaa apps soon to come.. 😁

zeke:
What about a hybrid layout with wall of icons for featured apps, then a filterable list of all the apps?

Kind of like in the blog post roundup:

screen shot 2017-01-11 at 9 35 11 pm

PR #654, opened 2017-01-13
[WIP] Apps List

zeke:
This is a work in progress. Has some funky styles that need to be cleaned up.

cc @simurai 

screen shot 2017-01-13 at 2 41 41 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,13 @@ env:
 branches:
   only:
     - gh-pages
+
+addons:
+  apt:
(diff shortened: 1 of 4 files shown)


zeke:
Bug:

screen shot 2017-01-13 at 2 47 32 pm

simurai:
Made a few changes. Can be seen here: http://simurai.com/electron.atom.io/apps/

zeke:
Nice!

PR #655, opened 2017-01-19
Add focus styles to app list

simurai:
This is a small follow up to #654.

After typing a search query, some people might just use `tab` to move down the list and then hit `enter`. This PR makes hover and focus styles the same. Border got added to make it a bit stronger.

music

👀  [Preview](http://simurai.com/electron.atom.io/apps/)


zeke:
👌

PR #659, opened 2017-01-24
Use EOL anchor when replacing links

kevinsawicki:
Looks like http://electron.atom.io/ is still showing 1.4.14 even though 1.4.15 is out.

I ran `./script/release` locally and the specs were failing on 1.4.15 because of the hyperlink spec, looks like some releases use `\r\n` and others `\n`.

This pull request just use the `$` anchor so both are handled.

### DIFF
--- a/script/releases
+++ b/script/releases
@@ -27,8 +27,8 @@ request(options, function (error, response, body) {
 
       // turn PR references like #123 into hyperlinks
       release.body = release.body.replace(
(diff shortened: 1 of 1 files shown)


zeke:
Thanks, @kevinsawicki. Opened https://github.com/electron/electron.atom.io/issues/660 as a reminder to make the release script fail more noticeably.

PR #666, opened 2017-01-31
removed stale, duplicated documentation

### LINKED ISSUE #658: <webview> Tag listed twice
FYI!

documentation_-_electron

shiftkey:
Fixes #658 

The canonical page over in `electron/electron` is named `webview-tag.md`:

And you can see the stale document here is named differently:

screen shot 2017-01-31 at 8 52 19 am

Tested the site locally, both places are now only showing one entry for `<webview>`.


MarshallOfSound:
Looks good 👍 

/cc @zeke Maybe the [`docs`](https://github.com/electron/electron.atom.io/blob/gh-pages/script/docs) script should be modified to wipe out the old docs before pulling in new ones (even if this is a rare occurrence)

zeke:
Thanks, @shiftkey! 

> Maybe the docs script should be modified to wipe out the old docs before pulling in new ones (even if this is a rare occurrence)

☝️  Opened #667 as a follow-up.

PR #670, opened 2017-02-03
Deployment Notifications

### LINKED ISSUE #660: Send a Slack notification when release script tests fail
The release script currently fails quietly when `npm test` doesn't pass. This means the website doesn't get updated automatically until humans notice that it's fallen behind. See https://github.com/electron/electron.atom.io/pull/659

To work around this, I think we should have a slack notifier thing that gets run when the tests fail.

Maybe https://github.com/andrewchilds/slack-notify would do the trick.

zeke:
This is an attempt to make the website's automated deployment script more vocal about its outcomes. I started out trying to implement a catch-all in bash, but that's not easily done.

- [x] Rewrite the script in `shelljs`, which we "can use it to eliminate our shell script's dependency on Unix while still keeping its familiar and powerful commands."
- [x] Rename `release` to `deploy` (to avoid potential confusion with unrelated `releases` script)
- [x] Send a failure message to Slack when tests fail or an uncaught exception is raised.
- [x] Send a success message to Slack when a deployment occurs.
- [ ] Change the Heroku Ccheduler command to `npm run deploy`

Fixes #660


zeke:
I'm reluctantly calling this "Ready for Review". Will probably take a few live trials to make sure it's working perfectly.

zeke:
Closing this in favor of a more Unix-y approach: https://github.com/electron/electron.atom.io/pull/672

PR #689, opened 2017-02-23
Glossary should be on website too.

sebcode:


zeke:
Nice. Thank you, @sebcode

PR #673, opened 2017-02-09
Always deploy docs for latest stable version of Electron

zeke:
Fixes #669 

This PR:

- Updates the script that fetches the GitHub API releases to sort them by version.
- Adds tests to verify that the release data never includes prereleases or drafts, and is in semver order, descending.
- Updates the docs build to get `highestVersion` from GitHub data, rather than the S3 data which includes _all_ Electron versions.

PR #696, opened 2017-03-01
use highest stable version when building docs

zeke:
I missed an important spot when implementing https://github.com/electron/electron.atom.io/pull/673 -- this is a fix for that.


zeke:
Effects of this change: https://github.com/electron/electron.atom.io/commit/b3fbb5cff7cd6d0bf465a73fa6c1096008605c46

PR #701, opened 2017-03-05
Added Designer.io Application

db2k:


db2k:
Should move it down in the array again since I renamed it to `Gravit Designer` instead of `Designer`.

Will fix that later.

50Wliu:
Apps need to be added at https://github.com/electron/electron-apps#adding-your-app, not here.

db2k:
@50Wliu I see, thanks!

PR #700, opened 2017-03-05
Filtered appslist, new featured app and counter for apps in _apps.json

db2k:
# This Pullrequest adds

### Counter on top of apps list with the total amount of apps
image

### Sorted applist by publishdate and reversed for loop
image

PR #704, opened 2017-03-06
App date improvements

zeke:
This is a followup to #700 that makes a few adjustments.

Relative times like "Added 3 days ago" are easier to understand than dates because you don't have to ask yourself "what is the date today?" then do subtraction. Also in this case the exact date that an app was submitted is not particularly important.

Having all the dates visible made things feel visually noisy, so I revised the CSS to only show dates when the row is active/hovered.

screen shot 2017-03-06 at 1 00 19 pm

cc @db2k


db2k:
Great one!
I wasn't sure how to do that with Jekyll but wanted to add that feature for now.

Looking forward to see that live! 👍

db2k:
@zeke one note for mobile:

I don't know how important the mobile design is, since I don't know the analytics for that, but I would propose to use a different way to layout the list and use table-row/table-cells or flexbox instead of inline elements with a set line-height.

That way we can break down the columns on mobile.
What do you think?

zeke:
> That way we can break down the columns on mobile.

Okay with me. cc @simurai who knows about the underlying CSS frameworks used to build this site.

simurai:
On mobile, currently the description just gets truncated more once you tap on an item:

screen shot 2017-03-07 at 10 52 32 am

Not sure about Android. You can test here: http://simurai.com/electron.atom.io/apps/

@zeke Do you have the analytics of the `/apps` page desktop vs. mobile? Yeah, if mobile gets used quite a bit we could consider wrapping the text or have two lines so.

db2k:
On Android the hover state is triggered by holding down the thumb on the a elements so I would rather switch to showing the addition date always and breaking the columns. I could come up with a solution.

@simurai which frontend framework is used?

simurai:
It's BaseCoat, I think originally just used for internal sites, but now part of http://primercss.io. It probably got changed quite a bit by now. We don't use it as a dependency and have it included, see the files starting with [_basecoat](https://github.com/electron/electron.atom.io/tree/gh-pages/_sass).

In other words, we could change anything we like. 😄

db2k:
Thanks, I'll change the markup for the application list a bit to make the layout better on mobile. Right now you can't really read the description anymore :)

PR is coming in around 3-4 hours, can't hop on right now

db2k:
Maybe we can also use a card layout with 50/50 columns?

iphone 7

Issue #705, opened 2017-03-07
Better /apps/ Layout on Mobile Devices


bdbch:
Regarding the mobile device layout I would propose a card layout for mobile devices. That's a pretty fast draft:

![image](https://cloud.githubusercontent.com/assets/6538827/23669865/fd4d1602-0365-11e7-94ee-936051927018.png)

Any thoughts about that?
Otherwise we could also just break down the description and date below the app title and keep the logo on the left.

simurai:
> Otherwise we could also just break down the description and date below the app title and keep the logo on the left.

This might be the better option when showing lots of apps. You can scan the list instead of going "zig-zag". The cards pattern would work great if there are just a few apps.

db2k:
I'll send in a pull request later. I agree jumping across the screen could be really displeasing on a large list.

👍

db2k:
Also just found this bug:
![image](https://cloud.githubusercontent.com/assets/6538827/23767255/425cc65e-0508-11e7-9757-1593a7d75dfa.png)

I'm working on a Pullrequest.

zeke:
@db2k what OS and browser are you using?

cc @ediweissmann who just submitted that app.

https://github.com/electron/electron-apps/pull/126

db2k:
**OS:** Windows 10 64 Bit Build 14939
**Browser:** Google Chrome 56.0.2924.87

ediweissmann:
I can confirm I also see the issue with hover state. Other apps in the list have same issue.

Some icons are very small on hover (as in screeenshot), some change size to a slighly smaller icon, some don't change icon size on hover.

Let me know if I can help with this.

zeke:
I just tried reproducing this issue in Google Chrome on my Windows 10 VM, but I don't see any problems. If one of you is able to submit a fix for this, it would be much appreciated. 👍

ediweissmann:
I opened a PR for the "small icon on hover state" thingy. https://github.com/electron/electron.atom.io/pull/711

zeke:
We are working on [converting the site from Jekyll to Node](https://github.com/electron/electron.atom.io/pull/758). Once we've got that shipped, we can really focus on improving the apps pages on the site. @jkleinsc has opened a new issue to discuss new design ideas for app pages: https://github.com/electron/electron.atom.io/pull/758

bdbch:
Sounds great! Looking forward to see the project in node.

zeke:
[...]

PR #711, opened 2017-03-10
re #705: Prevent app icon from scaling down when screen width is small.

ediweissmann:
In #705 is mentioned a small bug related to the app icon size on hover, on narrow screens.
https://github.com/electron/electron.atom.io/issues/705#issuecomment-285457761

This fixes it by not allowing the app icon to scale down anymore.


zeke:
Thanks!

PR #720, opened 2017-04-07
update icon of Caret

erusev:


zeke:
Thanks, @erusev. Would you mind opening a PR over on https://github.com/electron/electron-apps instead?

erusev:
Done, thanks!

PR #723, opened 2017-04-20
Updated Microstockr icon

paul321:


MarshallOfSound:
@paul321 This directory is auto generated, please PR into https://github.com/electron/electron-apps

PR #736, opened 2017-05-19
prevent tabnabbing by remove target _blank from all links

### LINKED ISSUE #732: Reverse tabnabbing fix is messing up referral analytics
We've heard from some Electron app developers that their referral traffic from the Electron website has plummeted in the last month. It seems that #718 is the cause of the issues, as the `rel="noreferrer"` attribute disables the Referer HTTP header.

I think we can solve this by removing `noreferrer` from the `rel` attributes of our links, and leaving `noopener` in place.

cc @sarabreeding who reported this issue.
cc @CodeTheory who confirmed a drop in traffic for Kap.
cc @EdOverflow who submitted the original fix.

zeke:
Fixes #732 

cc @sarabreeding @bnb @CodeTheory @EdOverflow

### DIFF
--- a/_includes/listed_app.html
+++ b/_includes/listed_app.html
@@ -1,6 +1,6 @@
 {% unless app.disabled %}
 <li class='listed-app'>
-  <a href='{{ app.website | default: app.repository }}' target='_blank' rel='noopener noreferrer' title='{{ app.name }} - {{ app.description | escape }}'>
+  <a href='{{ app.website | default: app.repository }}' target='_blank' title='{{ app.name }} - {{ app.description | escape }}'>
     <div class='listed-app-logo-wrapper'>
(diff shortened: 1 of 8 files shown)


_includes/listed_app.html
```diff
@@ -1,6 +1,6 @@
 {% unless app.disabled %}
 <li class='listed-app'>
-  <a href='{{ app.website | default: app.repository }}' target='_blank' rel='noopener noreferrer' title='{{ app.name }} - {{ app.description | escape }}'>
+  <a href='{{ app.website | default: app.repository }}' target='_blank' title='{{ app.name }} - {{ app.description | escape }}'>
```
CodeTheory: Does it not matter to remove `target='_blank'` here?

zeke: Ah I missed those on account of the single-quote. Fixed. Thanks.

PR #737, opened 2017-05-22
Fix missing open bracket in quick start title

jbw91:
Was browsing the site when I noticed the Quick Start title wasn't loading properly. Noticed it was missing the opening brace so I fixed it!


jbw91:
Just noticed that the `{{ site.data.locale.need_help.description }}` had template strings in it, which weren't interpolating. I fixed those to be consistent with that file, which had all links and such hard-coded.

zeke:
🙏 

Shipped that a bit hastily today. I'm eager to get a localization process in place!

PR #745, opened 2017-06-22
[Update] The logic of a sentence

zombieleet:
**Set an extra parameter to set be sent with the crash report** was corrected to

**Set an extra parameter to be sent with the crash report**


zeke:
Thanks @zombieleet! The source content is actually in another repo, so I've opened a new PR with the fix at https://github.com/electron/electron/pull/9881

PR #748, opened 2017-06-23
Run lighthouse on Travis

kuychaco:
This PR introduces a script to run Google Chrome's [lighthouse](https://github.com/GoogleChrome/lighthouse), a tool that audits web apps and web pages to help developers identify and fix common problems that affect performance, accessibility, and UX. The script will run on Travis and serve as a linter of sorts to catch regressions that might be introduced as the code evolves. The [`lighthouse.json`](https://github.com/electron/electron.atom.io/blob/9940031816538dc2b0bfdf78fff24a18a13fb517/lighthouse.json) file indicates which audits to run.

As part of this we built a command line tool called [`lightcrawler`](https://github.com/github/lightcrawler) that crawls a web site, runs the scraped links through lighthouse, and formats the output and provides a summary of the aggregated results. 

build__1523_-_electron_electron_atom_io_-_travis_ci


simurai:
Ohh... ✅  green!!!! 🙌 

Something I'm wondering: Is there a way to relax certain rules? Similar to linters. Because at some point we might want to merge something that should be an exception and then the build will fail forever.

kuychaco:
@simurai yeah that's a great question. We have a [`lighthouse.json`](https://github.com/electron/electron.atom.io/blob/08089f9ff4df2918e672c377c236988cb00d1ed5/lighthouse.json) file where you specify what audits you want to run. So you can just remove items from the `onlyAudits` field to relax the rule.

zeke:
This is awesome. Should we ship it?

kevinsawicki:
> This is awesome. Should we ship it?

We should, thanks @kuychaco and @simurai 👍 📈

simurai:
Also /cc @muan in case this is interesting.

tl;dr: You can hook up [Google's Lighthouse](https://github.com/GoogleChrome/lighthouse) to Travis, so that for every PR it checks against accessibility recommendations.

PR #770, opened 2017-09-11
add app.json for heroku

zeke:
Not actually needed for this branch, but Heroku requires it for Review Apps to work.


zeke:
screen shot 2017-09-11 at 8 59 13 am

PR #771, opened 2017-09-13
Add TouchBar and others TouchBar.. in Main Process list

Toinane:
Add TouchBar and others TouchBar in the main process list in /docs/
screenshot


MarshallOfSound:
I feel as though this folder is auto-generated from somewhere 🤔

Toinane:
Ah! I don't know where so? I've investigate this :) cc @zeke

zeke:
Nope this file is actually not generated. Curated by hand up until now, though it obviously doesn't need to be, given that we have `electron-api.json`

PR #773, opened 2017-09-28
add Chromium RCE blog post

zeke:


zeke:
https://electron.atom.io/blog/2017/09/27/chromium-rce-vulnerability-fix

PR #785, opened 2017-10-22
[WIP] Update links

codebytere:
updating relative links and links to electron.atom.io

### DIFF
--- a/data/awesome_electron.json
+++ b/data/awesome_electron.json
@@ -1714,12 +1714,12 @@
   },
   {
     "name": "Quick Start",
-    "href": "http://electron.atom.io/docs/tutorial/quick-start/",
+    "href": "http://electronjs.org/docs/tutorial/quick-start/",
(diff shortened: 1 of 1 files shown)


data/awesome_electron.json
```diff
@@ -1714,12 +1714,12 @@
   },
   {
     "name": "Quick Start",
-    "href": "http://electron.atom.io/docs/tutorial/quick-start/",
+    "href": "http://electronjs.org/docs/tutorial/quick-start/",
```
torn4dom4n: You should use `https://` instead of `http://`.

zeke:
@codebytere I'm gonna go ahead and close this out, as https://github.com/electron/electron.atom.io/pull/817 removes the awesome json file from the repo and gets data straight from the module.

At some point we should follow up and actually make the changes in the source at https://github.com/sindresorhus/awesome-electron

PR #788, opened 2017-10-23
move partials to views directory

zeke:


zeke:
Oops wrong source branch.

PR #793, opened 2017-10-25
All the Docs!

zeke:
This PR implements the `/docs/all` route which displays everything on a single page. This PR also localizes more of the content and improves docs index styles.

### DIFF
--- a/styles/_basecoat-page-headers.scss
+++ b/styles/_basecoat-page-headers.scss
@@ -59,25 +59,3 @@
     }
   }
 }
-
-
-// Minitron -----------------------------
-
-.minitron {
-  padding: $spacer3;
-  color: $mini-color;
-  background-color: $mini-bg-color;
-
-  a {
-    color: $mini-color-link;
-  }
-}
-
-.minitron-title {
-  margin: 0;
-  color: inherit;
-  font-size: 20px;
-  font-weight: 300;
-  letter-spacing: 1px;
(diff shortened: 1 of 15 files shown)


styles/_basecoat-page-headers.scss
```diff
@@ -58,26 +58,4 @@
...
-
-.minitron-title {
-  margin: 0;
-  color: inherit;
-  font-size: 20px;
-  font-weight: 300;
-  letter-spacing: 1px;
-  -webkit-font-smoothing: subpixel-antialiased;
-}
+}
```
codebytere: think u need a newline here :)

styles/_docs.scss
```diff
@@ -128,3 +128,8 @@
 
   width: 100%;
 }
+
+.docs p a,
+.docs li a {
+  text-decoration: underline;
+}
```
codebytere: here too

styles/_basecoat-page-headers.scss
codebytere: unless your atom config & our linter don't care about that in scss files 🤔

zeke:
>  you'll also need to delete that stuff from the route

Cool. Will do. We should make a habit of adding tests with every PR (as appropriate) so we'll know if we ever step on each others' toes.

codebytere:
sounds good to me! sorry about that :)

PR #796, opened 2017-10-27
Add link to the Code of Conduct

lee-dohm:
Add links to the Code of Conduct to the header and the footer to make it more visible to visitors to the site.

### DIFF
--- a/_data/locale.yml
+++ b/_data/locale.yml
@@ -8,7 +8,6 @@ nav:
   userland: Userland
   releases: Releases
   contact: Contact
-  conduct: Code of Conduct
 
 electron_is_easy:
   title: It's easier than you think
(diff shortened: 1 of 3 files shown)


lee-dohm:
@zeke I think I did put it in the footer nav (in addition to the header), unless I'm misunderstanding? We could take it out of the header and just leave it in the footer if you think that would be best :+1:

As to what brought this up, someone mentioned (though I don't recall where now) that they didn't know about the Code of Conduct because it wasn't mentioned in the documentation on the website. They didn't know that `CODE_OF_CONDUCT.md` was a thing ... so I figured they wouldn't be the only ones that don't know about that convention.

zeke:
> We could take it out of the header and just leave it in the footer if you think that would be best 

Yes please :)

PR #798, opened 2017-11-07
Catch Not Found pages of /docs and add title for categories of docs. (/docs/404)

pierreneter:
Visit: http://electronjs.org/docs/404
On localhost:
404-catch

This PR will:

- Add the title for docs's categories.
- Catch 404 pages.


routes/docs/category.js
```diff
@@ -1,26 +1,48 @@
...
+          title: `${selectedCategory.name} | Electron`
+        },
+        layout: 'docs',
+        processes: processes,
+        category: selectedCategory.name
+      })
+    }
+    res.render(`docs/${category}`, context)
+  }  else {
+    next()
```
zeke: I would prefer an early return at the top, for readability:

```js
if (!selectedCategories.length) return next()
```

Also, can you add a test(s) for this? See test/index.js

PR #799, opened 2017-11-07
Catch 404 pages of path: /userland/*

pierreneter:


routes/userland/show.js
```diff
@@ -1,33 +1,40 @@
...
+      items: report.collection,
+      blacklist: blacklist,
+      page: {
+        title: `${report.title} | Electron`
+      }
+    })
 
-  res.render('userland/show', context)
+    res.render('userland/show', context)
+  } catch (e) {
```
zeke: Rather than catching this entire block of code, can we just detect a 404 earlier and then `return next()`?

package.json
```diff
@@ -19,7 +19,7 @@
     "build-awesome": "script/awesome",
     "build-userland": "script/userland",
     "release": "script/release 2>&1 | script/notify",
-    "dev": "nodemon -e js,scss,yml server.js",
+    "dev": "NODE_PATH=. nodemon -e js,scss,yml server.js",
```
zeke: What is this `NODE_PATH` thing for?

server.js
```diff
@@ -7,11 +7,11 @@ const lobars = require('lobars')
 // Middleware
 const hbs = require('express-hbs')
 const slashes = require('connect-slashes')
-const browsersync = require('./middleware/browsersync')()
+const browsersync = require('./middleware/browsersync')
```
zeke: looks like some unrelated changes ended up in this file?

pierreneter: Yep. I just edit this for consistency in the `require`s. Remove `()` at this line and add to line 55.

package.json
pierreneter: Add `NODE_PATH` we can quickly `require` anything in project with a `path` no need to think about  not absolute path. Like I can do here: [routes/userland/show.js#line6](https://github.com/electron/electron.atom.io/pull/799/files/3b56398c6cb4e84666a4c8e1da429ee44b8efe6b#diff-6606e56f8941159c4982e8790c9bcef0R6)

PR #800, opened 2017-11-07
Add title for main categories

pierreneter:


pierreneter:
I don't know its existence in that `middleware`, that strings get picked up automatically but I don't see that work on my PC because of not title for this path don't add yet. So, I will add to this local file and commit again.
Sure, if this will be documented is great.

routes/home.js
```diff
@@ -2,7 +2,11 @@ const apps = require('../lib/apps')
 
 module.exports = (req, res) => {
   const context = Object.assign(req.context, {
-    apps: apps
+    apps: apps,
+    page: {
+      title: 'Electron | Build cross platform desktop apps with JavaScript, HTML, and CSS.',
```
zeke: This page title and description should live in data/locale.yml

pierreneter: I will move all it to locale.yml

PR #803, opened 2017-11-09
Stylize the language progress bars

MarshallOfSound:
Makes the progress bars look a bit fancier 😆  

image


zeke:
This change causes some kind of style regression in Firefox:

screen shot 2017-11-08 at 7 33 25 pm

zeke:
Looks great in Chrome though. 👍

MarshallOfSound:
@zeke Should be alg on firefox now 👍

PR #804, opened 2017-11-09
Add nvmrc

MarshallOfSound:
People using nvm will just have to run `nvm use` now to get the right node verison 👍


zeke:
Are you running on Node 9? I saw a failure on it earlier today and locked `engines.node` to '8' to work around it.

https://github.com/electron/electron.atom.io/commit/e4f05d87ac9b446ea45c075c717a3d09af94e078

MarshallOfSound:
Haven't even tried 9 yet, barely using 8 😆  Just anything <8 doesn't work for this and if this file is there my terminal automatically switches to the right node version 😀

PR #806, opened 2017-11-09
Blog update [WIP]

vanessayuenn:
- Update the blog section with the old look with excerpts to be more readable
- Some small changes to the blog postsl, since some formats used are not compatible with the new markdown parser we're using
- re-pin the primer-breadcrumb to a version that does not break the site (see primer/primer-css#317)

Some pending items:
- [ ] make author attribute in blog posts array-compatible
- [ ] implement new way to style tables in markdown


zeke:
> some formats used are not compatible with the new markdown parser we're using

Which ones?

zeke:
This looks awesome! Let's go ahead and ship it and tackle that short checklist in another PR(s).

PR #810, opened 2017-11-10
Fix Address of Korea Community

ssu2030:
fix address to "https://electron-kr.github.io/electron-kr"
I changed it to '@ragingwind' request.
You can refer to #11078 in electron/electron


zeke:
Thanks!

PR #813, opened 2017-11-10
Add subtron for /docs

pierreneter:


zeke:
If you rebase with master then the deployment should work. It was an issue where both yarn.lock and package-json.lock existed.

pierreneter:
Rebased and it works! 👍 👍 👍

PR #814, opened 2017-11-10
i18n for /community

pierreneter:


zeke:
This is great, @pierreneter! Thanks for all the PRs.

Would you be willing to add some basic tests for this page? Nothing fancy, just enough to know that the page is being properly rendered and displaying some of the localized content.

zeke:
Looks like this needs a rebase. Otherwise good to go.

PR #817, opened 2017-11-12
Clean up and test community page

zeke:
This PR removes the old awesome-electron.json file that Jekyll needed, and instead gets data straight from the awesome-electron module. This PR also localizes more content on the community page.


zeke:
> why didn't we require it instead of keeping data locally before?

because it used to be a ruby app! :)

PR #821, opened 2017-11-13
Nav styles

zeke:
This PR updates the nav to dynamically change color based on the colors of the current app you're viewing. All colors have their contrast ratio tested in electron-apps, so they're guaranteed to be accessible.

screen shot 2017-11-13 at 8 10 18 am

screen shot 2017-11-13 at 8 09 34 am

screen shot 2017-11-13 at 8 09 58 am

cc @donokuda @simurai


zeke:
An alternative way for Electron's branding/nav to "take a back seat" on these app pages,would be to make the nav background white with the Electron logo and nav links some faint grayish color.

Toinane:
Cool idea, what about logo with multiple colors in?

donokuda:
⚡️ Electrifying! ⚡️ 

This looks awesome. I'm not sure what the branding guidelines are for the website (and if changing the header color is okay), so this is something I'd like @simurai to chime in when he gets back.

That said, having the header match the custom link colors is  🌈 ✨ and feels more consistent overall as a result.

zeke:
> what about logo with multiple colors in?

@Toinane the most "bold" color in the logo is selected. The implementation for that is here: https://github.com/zeke/pick-a-good-color and here: https://github.com/zeke/make-color-accessible

zeke:
I'm gonna go ahead and ship this and If it's not well-received by the Electron maintainers group (or @simurai :)) we can roll it back.

Toinane:
Okay cool, Thanks for the details :)!

PR #822, opened 2017-11-13
[WIP] Polish app details page

donokuda:
Hi! Hello! Yes this is Don!

This pull request is an attempt to style the app pages closer to what we have for [GitHub Marketplace](https://github.com/marketplace/coveralls). My hope is that it's Good Enough™ that it doesn't look like a knock-off, but if I were honest with everyone, it sort of feels like it because the styles aren't mapped 1-to-1.

Either way, I still feel like it's an improvement so here we go! 🤓 

Fun fact: I kept flipping back and forth on whether the README should go before or after the screenshots. Finally settled on having it before since it's unexpanded anyway and therefore doesn't originally take up a lot of space.

🔙 **Before (README unexpanded)**

localhost_5000_apps_todu 3

🔙 **Before (README expanded)**

localhost_5000_apps_todu 4

🔜 **After (README unexpanded)**

localhost_5000_apps_todu 1

🔜 **After (README expanded)**

localhost_5000_apps_todu 2

/cc @simurai @codebytere @zeke for 👀 and 💭


donokuda:
@zeke I need some help on one part of this redesign. I'm trying to create a gallery-like experience for the screenshots, but I'm unsure how to JavaScript 😕 



I've tried adding something as simple as a `console.log("sup don");` inside the little `<script>` tag  
but it doesn't seem to appear in the dev tools.

https://github.com/electron/electron.atom.io/blob/572cc6f62ab4be3525b4aee68d924b1bbd94a4b6/views/apps/show.html#L163-L168

The functionality I'm trying to recreate can be seen on any marketplace app details page: (For example: https://github.com/marketplace/coveralls)

zeke:
Yay! This is looking really nice, @donokuda. Thanks for jumping in. I can help make the screenshot gallery thing work. I'll open a PR against this one.

zeke:
The default branch on this repo was just changed from `neo` to `master` and GitHub doesn't allow post-facto changes to a PR's target branch, so I'm going to have to close this and open a new PR.

PR #828, opened 2017-11-14
Help section now links to correct twitter account

Kilian:
I looked elsewhere and could not find a mention of twitter_username, but there are a couple of direct links to twitter.com/electronjs (just search this repo for "twitter"), so I opted to fix it that way (as the other links in this section are direct links too)

### DIFF
--- a/data/locale.yml
+++ b/data/locale.yml
@@ -109,7 +109,7 @@ quick_start:
 
 need_help:
   title: Need Help?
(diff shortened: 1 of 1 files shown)


Kilian:
oh, and here's the image of the current homepage: selection_001

Kilian:
...As per the above deployment, I need to do something more before it's actually applied to the homepage, and that's in the i18n repo/workflow. The problem actually occurs in all languages:

selection_002

and these files should be updated:
https://github.com/electron/electron-i18n/search?p=3&q=twitter_username&type=&utf8=%E2%9C%93

vanessayuenn:
I believe once the fix is merged into `locale.yml` in the neo branch on this repo, it should be picked up and updated in all the other i18n locales. @zeke will definitely have a more concrete answer on that.

zeke:
Thanks, @Kilian. 👍 

Like @vanessayuenn said, the new string will be displayed on the site in English for all locales until it goes through the translation cycle. See https://github.com/electron/electron.atom.io/blob/neo/CONTRIBUTING.md#localized-strings for more info on how that works.

PR #833, opened 2017-11-15
Hide app categories on mobile

codebytere:
This PR hides app categories on mobile, but leaves filter and moves it above the app list. 

screen shot 2017-11-15 at 3 43 39 pm

screen shot 2017-11-15 at 3 43 48 pm

/cc @zeke


zeke:
Fantastic. Thank you for this!

PR #846, opened 2017-11-16
enable community notice

zeke:


zeke:
`hub` still thinks gh-pages is the default branch. 🤔

PR #849, opened 2017-11-16
electron.atom.io → electronjs.org

zeke:
This PR updates a bunch of references to the old domain.

cc @codebytere


zeke:
Hmm. Got this from Travis:

> Done: Job Cancelled
> 
> This job was cancelled because the "Auto Cancellation" feature is currently enabled, and a more recent build (#2350) for pull request #849 came in while this job was waiting to be processed.

PR #856, opened 2017-11-16
disable email notifications from travis

zeke:


zeke:
Oops wrong target branch wtf

PR #858, opened 2017-11-16
update electron-apps

zeke:


zeke:
Not sure why @greenkeeper didn't do this..

PR #870, opened 2017-11-18
Don't auto execute shell one liners

MarshallOfSound:
Currently if you copy paste a one liner from the site into terminal it will copy a newline char as well, this will cause the one liner to automatically execute.  My terminal won't auto exec but it does throw warnings my way whenever it prevents something auto executing (how I noticed this) 😄


zeke:
👍

PR #873, opened 2017-11-19
Fix 301 redirected links

him2him2:
Update Links to reduce 301 redirections for faster URL resolution.


zeke:
Great idea! Thanks for the fix.

PR #874, opened 2017-11-19
refactor electron.atom.io to electronjs.org

him2him2:
replaceing links and fixing redirects:
electron.atom.io -> electronjs.org


MarshallOfSound:
Hey @him2him2,

All these files are automatically generated from other data sources.  I don't think we should patch them manually rather the source of the information should be updated.

/cc @zeke RE the homepage of the electron package

zeke:
Yeah this data comes from other sources. Pretty sure I got everything we need in https://github.com/electron/electronjs.org/pull/849

PR #883, opened 2017-11-22
gzip-encode response

groundwater:
I tried running lighthouse against the website, and it recommends gzipping the responses

image

Implemented something, as recommended by [express](https://expressjs.com/en/advanced/best-practice-performance.html#use-gzip-compression). Alternatively they recommend using a reverse proxy.


zeke:
⚡️

PR #885, opened 2017-11-22
Lazyload-app-icons edits

sethlu:


scripts/lazy-load-images.js
```diff
@@ -1,20 +1,56 @@
...
-      }
-    })
-}
\ No newline at end of file
+  ['scroll', 'resize', 'load'].forEach(event =>
+    window.addEventListener(event, updateImages)
+  )
+
+  // add all images that are part of the DOM tree
+  // NB: |display: none| elements need manual adding
```
sethlu: Related issue with the list items should be addressed in [@9e619ce/scripts/create-filter-list.js#L33](https://github.com/sethlu/electronjs.org/blob/9e619ce4621d5b366b43efd7fbe78814baf372ac/scripts/create-filter-list.js#L33), and slightly optimized with with an extra `image.dataset.src` check in `addImage(image)` so a same image is not requeued.

scripts/lazy-load-images.js
```diff
@@ -1,20 +1,56 @@
...
+  image.onload = () => {
+    image.removeAttribute('data-src')
+
+    // remove colored circular placeholder element
+    const placeholder = image.parentElement.querySelector('.placeholder')
+    if (placeholder) placeholder.classList.add('faded-out')
+  }
+}
+
+// TODO: Read/write lock on |images|?
```
zeke: Can you clarify what this TODO mean?

scripts/lazy-load-images.js
```diff
@@ -1,20 +1,56 @@
...
+let updateImages = module.exports.updateImages = throttle(() => {
```
zeke: This `let` can be a `const`

scripts/lazy-load-images.js
```diff
@@ -1,20 +1,56 @@
...
+
+let addImage = module.exports.addImage = function (image) {
+  // abort if this element has previously entered the viewport
+  if (!image.dataset.src) return
+
+  images.push(image)
+  updateImages()
+}
+
+function handle (image) {
```
zeke: Can you use a more descriptive name for this method, like `displayImage` or similar?

scripts/lazy-load-images.js
```diff
@@ -1,20 +1,56 @@
...
+  )
+
+  // add all images that are part of the DOM tree
+  // NB: |display: none| elements need manual adding
+  document.querySelectorAll('img[data-src]').forEach(addImage)
+}
+
+let images = []
+
+let addImage = module.exports.addImage = function (image) {
```
zeke: This `let` can be a `const`

scripts/lazy-load-images.js
sethlu: Since I applied a filter over the list, I am thinking if there will be cases when `addImage` is called (adding another element into the list) while the loop is taking place.

Do we want to have:

```
let copy = images;
images = []
// test & filter elements in copy
// move elements from images (probably not empty because `addImage` is called)
// images = copy
```

sethlu:
@zeke Let me know if there are other necessary changes. 😺

zeke:
Thanks, @sethlu!

PR #895, opened 2017-11-26
Fix before of hljs code class

### LINKED ISSUE #859: Code tags' content is shifted to the left
code-tag-content-shifted-left

These styles seem to be the culprit:
code-tag-styles

pierreneter:
Fix #859


styles/hljs/overrides.scss
```diff
@@ -2,9 +2,8 @@
   line-height: 1.6;
   font-size: 1.2em; 
   padding: 10px;
+  &::before {
```
zeke: Shouldn't this just be a single colon, like ` &:before` ?

50Wliu: It should be `::before`, which is the preferred syntax added in CSS3.  Though `:before` definitely works as well (and is better for backwards-compatibility).

https://developer.mozilla.org/en-US/docs/Web/CSS/::before#Syntax

zeke: Today I Learned™ :)

PR #911, opened 2017-11-29
Add category slug and call the view README button back to work

pierreneter:


zeke:
@pierreneter can you summarize the problem this is solving?

pierreneter:
@zeke sure, 
- Accessing an app detail will display the app's category (if any). Add `categorySlug` of this app's category to enable to click back to apps list with this category be selected.
- The view README.md of an app in app detail page not work, I just uncomment this code to call it back to work.

zeke:
Nice. Thank you!

PR #913, opened 2017-11-29
move logic into one place, avoid the 404 case appears suddenly

### LINKED ISSUE #909: 404 at https://electronjs.org/apps but the page seems to be loaded correctly
Title of https://electronjs.org/apps says '404 not found', but the content is loaded correctly

pierreneter:
Try to fix #909, make it never happen anymore


pierreneter:
@zeke
> What is the purpose of the titled boolean? It doesn't appear to be used anywhere else, and this handler should only be called once per request.

About `titled`, because of I used: `req.context.page.title = `${req.context.page.title} | Electron`
After app started. The title if it seted it with be set again, example:

- First visit: `Title | Electron`
- F5: `Title | Electron | Electron`
- F5: `Title | Electron | Electron | Electron`

The boolean keeps the first title intact.

> Also we should probably be using the localized req.context.localized._404.page_not_found reference here instead of a handwritten English title.

Agree

> Lastly, if there is no page metadata, that doesn't necessarily mean that the page itself is not found. It just means that there's no page metadata for the given URL. Does that make sense?

Sure, but this is top middleware, this 404 Not Found will be the default for the page with empty info set, of course. (for top static pages). And routes behind can be set title anyway

zeke:
> Title | Electron | Electron | Electron

Ah I didn't realized that was happening. This is a sign that we have some state that's getting mutated and persisted across requests. I think this is the offending line:

```js
const page = i18n.website[req.language].pages[req.path] || {}
```

We should change it to this:

```js
const page = Object.assign({}, i18n.website[req.language].pages[req.path])
```

That way we'll get a copy of the page data without mutating it. This should fix the mysterious 404 error and remove the need for the `titled` boolean.

pierreneter:
Ok, I will change this code

middleware/context-builder.js
```diff
@@ -8,9 +8,17 @@ module.exports = function contextBuilder (req, res, next) {
   // Attach i18n object to request so any route handler can use it if needed
   req.i18n = i18n
 
+  const localized = i18n.website[req.language]
+
   // Page titles, descriptions, etc
-  const page = i18n.website[req.language].pages[req.path] || {}
-  page.path = req.path
+  let page = Object.assign({
+  	title: `${localized._404.page_not_found} | Electron`,
```
zeke: There will be some cases where the page is found doesn't have a title in locale.yml. Instead of setting the title to 404 there, the title should just default to something like 'Electron'. Does that make sense?

pierreneter: Ok

zeke:
Looks like this needs a rebase, as a few PRs have since landed on master.

zeke:
Nice! Hopefully this is squashed for good. Thanks for your persistence. 👍

PR #916, opened 2017-11-29
Add feature: copy code to clipboard

### LINKED ISSUE #871: Add buttons to copy code snippets to clipboard
https://clipboardjs.com/

pierreneter:
Close #871


zeke:
This is fantastic! I have a feeling this little feature will make people happy. :)

I noticed that the text becomes selected after clicking the copy button:

[screen shot 2017-11-29 at 2 36 02 pm](https://electron-website-pr-916.herokuapp.com/docs/api/browser-window#setting-backgroundcolor)

 I didn't really expect that to happen when I clicked. I guess it's good because I can now <kbd>cmd</kbd><kbd>c</kbd> if I'm paranoid that it didn't really copy, but to me the `Copied` tooltip feels like enough to give confidence that the code is actually on my clipboard.

What do you think?

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# editorconfig.org
```
zeke: Can you remove this file? Maybe it should be gitignored?

package.json
```diff
@@ -27,20 +27,21 @@
...
     "browserify-middleware": "^8.0.0",
+    "clipboard": "^1.7.1",
     "compression": "^1.7.1",
     "connect-browser-sync": "^2.1.0",
     "connect-slashes": "^1.3.1",
     "cookie-parser": "^1.4.3",
     "cross-env": "^5.1.1",
     "dotenv": "^4.0.0",
-    "electron-apps": "1.1976.0",
+    "electron-apps": "^1.1976.0",
```
zeke: can you revert these version updates that aren't related to the PR? That will help avoid merge conflicts, as electron-apps was just updated in a more recent PR that landed.

scripts/copy-code-to-clipboard.js
```diff
@@ -0,0 +1,26 @@
...
+    let id = `_${Math.random().toString(36).substr(3)}`
+    code.id = id
+    let copyButton = document.createElement('button')
+    copyButton.classList.add('btn-clipboard', 'simptip-position-top')
+    copyButton.setAttribute('data-tooltip', 'Copy to clipboard')
+    copyButton.setAttribute('data-clipboard-target', `#${id}`)
+    copyButton.innerHTML = 'Copy'
+    code.parentElement.appendChild(copyButton)
+  })
+  document.querySelectorAll('.btn-clipboard').forEach(button => {
```
zeke: Using Cliboard's event handlers might make this a little cleaner: https://github.com/zenorocha/clipboard.js#events

Also it would mean we'd have a `clipboard` instance and could get rid of the `// eslint-disable-line` comment.

.editorconfig
pierreneter: Yes. But I think it's useful?

package.json
pierreneter: I ran the command `npm install` (or `npm update` I don't remember) and it automatically added this. I will remove this change.

scripts/copy-code-to-clipboard.js
pierreneter: This `disable` for `Do not use 'new' for side effects.` :(

.editorconfig
zeke: It is useful, but we generally just use [standard](http://ghub.io/standard) as our opinionated linter of choice on Electron projects.

scripts/copy-code-to-clipboard.js
zeke: I know! I just meant you could do this:

```js
const clipboard = new Clipboard('.btn-clipboard')

clipboard.on('success', (event) {
    console.info('Action:', e.action)
    console.info('Text:', e.text)
    console.info('Trigger:', e.trigger)
    e.clearSelection()
})
```

.editorconfig
pierreneter: I don't see that in dev mode? Can we add `eslint` for that word on local?

zeke: npm test: `NODE_ENV=test mocha && standard --fix`

pierreneter: Thanks! I will update this PR's branch to update this scripts.

zeke:
@pierreneter I made some tweaks. Let me know what you think!

pierreneter:
LGTM! 👍 now it works smoothly

Issue #910, opened 2017-11-29
Generalize the Crowdin Proxy


zeke:
Blocking https://github.com/electron/electronjs.org/issues/867

There's a new endpoint https://electronjs.org/language-stats.json implemented in https://github.com/electron/electronjs.org/pull/903 that allows GET requests to Crowdin's project stats endpoint.

We should generalize this to allow arbitrary GET requests to the Crowdin API for the `electron` project. This will allow Electron contributors to develop new features that depend on Crowdin data without the need for a Crowdin API key. (Crowdin does not currently have the concept of a read-only key.)

## Example Mappings

Source | Target
------- | -------
electronjs.org/crowdin/language-status | api.crowdin.com/api/project/electron/language-status
electronjs.org/crowdin/status | api.crowdin.com/api/project/electron/status

Any existing query params in the request should preserved, and the following two query params should be injected:

```js
{
    key: process.env.CROWDIN_KEY,
    json: 'true'
}
```

https://api.crowdin.com/api/project/{project-identifier}/download-glossary?key={project-key}

## Performance / Abuse 

If we ever start to see too many requests to these endpoints, or if we go over our Crowdin rate limit, we can implement some in-memory or on-disk caching strategy. However for the short term I don't think caching will be necessary.

## Example Crowdin APIs for reference

- https://support.crowdin.com/api/info/
- https://support.crowdin.com/api/status/
- https://support.crowdin.com/api/language-status/
- https://support.crowdin.com/api/download-glossary/

Issue #867, opened 2017-11-17
Add a "Translate this Doc" link to docs footer


zeke:
At the bottom of each doc page, we have these links:

<img width="590" alt="screen shot 2017-11-17 at 3 30 28 pm" src="https://user-images.githubusercontent.com/2289/32973431-4437ebdc-cbac-11e7-9958-5e90bd0335f9.png">

We should add another that says "Translate this Doc".

If the visitor's locale is the default English (US, Canadian, or otherwise), one of the following should happen:

a) link to Crowdin project landing page. This is very easy to implement but is not the best UX, as the person has to then pick a target language and re-navigate to the file in question:

b) display a modal (or a new page, or a `<select>` element) that prompts for desired target langauge, then redirects to the Crowdin page for that language/filename pair on Crowdin.

If the visitor's locale is **not the default**, clicking "Translate this Doc" should take them straight to their language/filename pair on Crowdin. 🎯 

I've created a module [get-crowdin-file-ids](https://github.com/zeke/get-crowdin-file-ids#readme) that we weave into the output of the `electron-i18n` module, so we generate these "straight to the right translation page" URLs from the website.

zeke:
cc @pierreneter who is working on this.

zeke:
I moved the `get-crowdin-file-ids` repo to the @electron org, made it work with the new electronjs.org crowdin proxy, and set up semantic releases for it.

@pierreneter this should unblock you. 🚀

PR #923, opened 2017-11-30
🔧 fix source file link redirect on docs

codebytere:
Resolves https://github.com/electron/electronjs.org/issues/877

Add url to page context so that `propose change to source file` redirects appropriately to the docs page on `electron/electron`

/cc @50Wliu @zeke


zeke:
Looks good, @codebytere. Can you add a test so it doesn't regress again?

cc @pierreneter who may have also been working on a fix for this.

pierreneter:
Keep going! I'm doing #867, but it's still waiting for #910

PR #925, opened 2017-11-30
Update meetups's href at /community

pierreneter:
Update href for meetups and change to the correct link to pull a request.
capture4


views/community.html
```diff
@@ -81,10 +81,10 @@ <h2 id="meetups" class="mt-7">{{localized.community.meetups}}</h2>
       {{#each meetups}}
       <tr>
         <td>
-          <a href="{{ meetup.href }}">{{ this.name }}</a>
+          <a href="{{ href }}" target="_blank">{{ name }}</a>
```
zeke: A while back we opted to take out all the `target` attributes to prevent [tabnabbing](https://github.com/electron/electronjs.org/pull/736)

PR #927, opened 2017-11-30
Fix get selected nav

pierreneter:
capture10
capture11


scripts/apply-active-class-to-active-links.js
```diff
@@ -1,10 +1,13 @@
 module.exports = function applyActiveClassToActiveLinks () {
-  const topPath = location.pathname.split('/')[1]
+  const topPath = `/${location.pathname.split('/')[1]}`
 
   // homepage
   if (!topPath.length) return
 
-  document.querySelectorAll(`nav a[href^="/${topPath}"]`).forEach(a => {
```
zeke: Tricky little bug! 🐞 

I think we could solve this more simply by just changing `href^="/${topPath}"` to `href="/${topPath}"` (without the caret ^)

pierreneter: Hm... it's not the same with my logic. My way is to take the `a`'s `href` as a pattern to determine the exact path selected.

pic

zeke:
Thanks!

PR #929, opened 2017-11-30
add github and twitter icons to the footer

zeke:
Looks a little weird on desktop. Thoughts on how to make it look better, @donokuda?

## Macbook Pro 13

screen shot 2017-11-30 at 1 09 35 pm

## Mobile

screen shot 2017-11-30 at 1 09 56 pm

Maybe we could drop the "made with love" thing and replace it? After all, Electron is not just a GitHub joint anymore...


simurai:
> Maybe we could drop the "made with love" thing and replace it? After all, Electron is not just a GitHub joint anymore...

👍  Yeah, I think dropping the "made with love" would be a nice gesture towards all the other contributors and GitHub wouldn't look like it's trying to take all the credit.

How does it feel to replace "made with love" with the GitHub and Twitter icon?

donokuda:
I'm 👍 for removing the "made with love" as well and having the social links replace them. If we're adding the link to GitHub on the bottom, would it make sense to remove the link in the header?

image

zeke:
Okay I took out "made with love" and replaced it with icons linking to GitHub and Twitter.

> would it make sense to remove the link in the header?

@donokuda I'm open to it, but if we do I think we should consider replacing it with a text link to `GitHub`, so folks can still discover the GitHub org easily.

This is ready for 👀 again.

styles/_basecoat-footer.scss
```diff
@@ -13,17 +13,26 @@
 }
 
 .footer-nav {
+
```
donokuda: I'm totally being a stickler here, but did you want to 🔥 this blank line?

zeke: Fixed! This made me curious about Sass linting. Have folks on the design team experimented with style linting at all?

Currently seeing over 1000 issues using the vanilla [sass-lint](https://ghub.io/sass-lint) config: https://gist.github.com/zeke/bb7abcdab4a95ac702d4486cb6d65d42

donokuda: > Have folks on the design team experimented with style linting at all?

Yup! They have something for Primer that might be interesting for you: https://github.com/primer/primer/tree/master/tools/stylelint-config-primer

> Currently seeing over 1000 issues using the vanilla sass-lint config: 

I'd expect that to happen when introducing rules after the fact, which can be frustrating at times. Something to consider doing is disabling rules that you would not consider important. Another approach to introducing linting is something that my friend does with [Access Lint](https://www.accesslint.com/) is to report regressions (such as only lint code that has changed and not everything at once.)

zeke: Nice! I went ahead and set up [AccesLint](https://www.accesslint.com/) for this repo. Let's see how it goes.

PR #932, opened 2017-12-01
Add feeds

### LINKED ISSUE #890: Add a blog feed
used to be https://electron.atom.io/feed.xml

### LINKED ISSUE #931: RSS feed missing from new website
Hello!

It seems that new website and new blog don't have RSS feed. Is it intentional? Or is it something that will be added in the future? It was quite useful.

Thanks!

pierreneter:
Close #931
Close #890


routes/feed.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.exports = function feedHandler (req, res, next) {
+  let feed = new Feed({
+    title: 'Electron',
+    description: 'Build cross platform desktop apps with JavaScript, HTML, and CSS.',
+    id: 'http://electronjs.org/',
+    link: 'http://electronjs.org/',
+    generator: 'Electron website',
+    feedLinks: {
+      json: 'https://electronjs.org/feed.json',
```
zeke: Is `/feed.json` and `/feed.json` a convention? Or can we name it anything we want? If we can name it anything, I would prefer `blog.xml`, as that would free us up to have a pattern for things like `/releases.xml` in the future.

routes/feed.js
```diff
@@ -0,0 +1,37 @@
...
+      atom: 'https://electronjs.org/feed.xml',
+    }
+  })
+  req.context.posts.forEach(function (post) {
+    feed.addItem({
+      id: `https://electronjs.org${post.href}`,
+      title: post.title,
+      content: post.content,
+      description: `${striptags(post.content).split(' ').slice(0, 200).join(' ')} [...]`,
```
zeke: Maybe the `[...]` should only be appended to the string if there is actually truncated content?

middleware/blog.js
```diff
@@ -22,17 +22,24 @@ async function parsePost (filename) {
...
-  if (!req.path.startsWith('/blog')) return next()
+  const isFeed = ['/feed.xml', '/feed.json'].indexOf(req.path) > -1
+  if (!(req.path.startsWith('/blog') || isFeed)) return next()
 
   const context = Object.assign(req.context, {posts: posts})
 
   if (req.path === '/blog') {
     return res.render('posts/index', context)
   }
+  if (isFeed) return next()
```
zeke: Is this needed twice?

routes/feed.js
pierreneter: Yes, we can name it anything we want.

zeke:
I just gave the two XML and JSON feeds a try in Feedly and they look great! 👍 

Is RSS Autodiscovery still a thing? Wondering if we should add some meta tags in the head... http://www.rssboard.org/rss-autodiscovery

```html
<link rel="alternate" type="application/rss+xml" title="Electron Blog" href="https://electronjs.org/blog.xml" />
```

PR #937, opened 2017-12-05
Redirect docs/v0* to docs/*

pierreneter:
docs-v0


vanessayuenn:
I think at this point, we can probably drop the pre-1.0 docs redirect. The docs on the API now are way different from those of v0.1.1, I think a redirect for docs that outdated is more misleading than a 404. Thoughts on this @zeke?

pierreneter:
This redirect is only temporary. We will develop multi-version of docs for the website then this navigation will return as before.

PR #939, opened 2017-12-06
Proxy for some crowdin API endpoints

### LINKED ISSUE #910: Generalize the Crowdin Proxy

vanessayuenn:
Resolve #910 

This will allow access to certain Crowdin API endpoints without the project API key through `electronjs.org/crowdin/[xyz]`. For now, the endpoint whitelist is limited, but it can be extended if need be in the future. 

Outstanding items
- [x] get test to pass (proxy seems to be misbehaving when being created from test suite)
- [x] remove `/language-stats.json` endpoint and replace consumption of it to `/crowdin/status`


routes/languages/proxy.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+  if (!(apiWhitelist.find((path) => path === url.parse(req.url).pathname))) {
+    return res.status(404).render('404', req.context)
+  }
+  return proxy({
+    target: 'https://api.crowdin.com/api/project/electron',
+    changeOrigin: true,
+    logLevel: 'warn',
+    pathRewrite: (path, req) => {
+      const newPath = url.parse(path)
```
zeke: Rather than using `search` and concatenating strings, I think using `url.parse(path, true)` to preserve any existing query params is safer. That second argument parses the query string into an object. See https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost

Then:

```js
newPath.query = Object.assign(newPath.query, {
  key: process.env.CROWDIN_KEY,
  json: true
}
```

test/index.js
```diff
@@ -275,29 +275,47 @@ describe('electronjs.org', () => {
...
+      delete process.env.CROWDIN_KEY
+      const res = await supertest(app).get('/crowdin/status')
       res.statusCode.should.equal(401)
       res.type.should.equal('application/json')
       res.text.should.eq('"process.env.CROWDIN_KEY is not set"')
     })
+
+    test('returns 405 on request with method other than GET', async() => {
+      const res = await supertest(app).post('/crowdin/add-file')
+      res.statusCode.should.equal(405)
```
zeke: 405! Never used that one before. 👍

routes/languages/proxy.js
vanessayuenn: Ha I didn't know about the `parseQueryString` option

vanessayuenn:
Ready for another 👀 @zeke

PR #945, opened 2017-12-07
update meetup.json

dayanandhandaya:


zeke:
The link you've provided doesn't say anything about Electron. Can you explain?

Also, the JSON syntax is broken in your update.

zeke:
This doesn't appear to be an Electron meetup, so I'm going to close. Feel free to comment and update the URL if I'm mistaken. Thanks.

PR #948, opened 2017-12-08
Fix crowdin YAML parsing error by wrapping string in quotes

TiagoDanin:
# Problem

electron-i18n repo: 
image

https://github.com/electron/electron-i18n/blob/03a69d6695ad22feb05b99c25466c71b802261b9/content/pt-BR/website/locale.yml#L57

website used the electron-i18n: 
image

explore.description is converted to table in Crowdin

# Solution

Using a single string to explore.description

# Notify member

@zeke


pierreneter:
Wait for a second, I'm checking on this.

pierreneter:
Strangely, any like `the meetup_notice` field still works without `"`

https://github.com/electron/electron-i18n/blob/master/content/en-US/website/locale.yml#L43
https://github.com/electron/electron-i18n/blob/master/content/pt-BR/website/locale.yml#L31

TiagoDanin:
Because the first line does not have URL

See my test: [test.yml](https://gist.github.com/TiagoDanin/b37cb3254d47af1237198ca9512bc707)
image

zeke:
Ideally we wouldn't have to use quotes in situations like this, but this appears to be a bug in Crowdin's YAML parsing so this is a reasonable workaround for now.

Once CI is passing on this I'll merge it.

PR #952, opened 2017-12-13
Only one search input at /apps

### LINKED ISSUE #951: App search input is empty on mobile
Currently there are two identical search input elements on the /apps page. One is displayed in the sidebar on desktop, the other is displayed on mobile (or a narrow browser window). When viewing a URL like https://electronjs.org/apps?q=music on mobile, the input does not contain the existing search query.

I think the best approach to solving this is to only have on search element on the page (instead of two), and update the styles so that the input adjusts its appearance depending on context / window size.

pierreneter:
Close #951


styles/_app.scss
```diff
@@ -128,6 +128,13 @@
 // hide filter categories on mobile
 @media only screen and (max-width: 750px) {
   .categories-filter {
+    width: 100%;
+    padding: 0 !important;
```
vanessayuenn: Using `!important` generously can be a slippery slope. Can we achieve the same result without that?

styles/_app.scss
```diff
@@ -128,6 +128,13 @@
 // hide filter categories on mobile
 @media only screen and (max-width: 750px) {
   .categories-filter {
+    width: 100%;
```
vanessayuenn: Instead of specifying the width in the stylesheet here, we can just use `col-xs-12 col-md-3` in the template.

styles/_app.scss
pierreneter: I want to make it look like the current version. Keep the style that all of you have designed. Can compare at https://electron-website-pr-952.herokuapp.com/apps and https://electronjs.org/apps (browser width is small)

styles/_app.scss
pierreneter: No, it does not work, you can try. It's fixed with: `max-width: 750px`. `col-*` don't work anymore.

styles/_app.scss
pierreneter: Let see, what I do just to make should be nothing different from the current design

styles/_app.scss
vanessayuenn: Thanks for bringing that to my attention, I missed the `max-width: 750px`. We already have some breakpoints ([sample usage](https://github.com/electron/electronjs.org/blob/master/styles/_base.scss#L7)) defined with basecoat, so we shouldn't add another one at 750px. That way we can keep using `col-*` for consistency.

pierreneter: @zeke what do you think about refactoring the design of apps page to use `col-*`?

pierreneter:
@zeke done!

PR #960, opened 2017-12-15
add doc histories, serve release data, improve docs footer

### LINKED ISSUE #820: Serve structured Electron releases data
https://atom.io/download/electron/index.json is a redirect to https://gh-contractor-zcbenz.s3.amazonaws.com/atom-shell/dist/index.json, a file which maps Electron versions numbers to their corresponding versions of V8, Chromium, Node, etc.

We should serve this data from the website instead of S3, and make it an official™ endpoint, maybe `/releases.json` or `/versions.json`.

We should also denote the beta status of a release in this data.

cc @Kilian

zeke:
Apologies for this giant PR, which snowballed into a bunch of things. ☃️ 

Here are the highlights:

## Electron Releases Data

- Is now a public endpoint: https://electron-website-pr-960.herokuapp.com/releases.json
- Includes everything that the GitHub Releases API returns
- Now includes all (GitHub) releases, not just the last 100
- Folds in data from `https://atom.io/download/electron/index.json` like v8 version, node, chromium, etc in a prop called `dependencyVersions` (open to alternative names for that prop)
- Is 5.6MB (is that a problem?)

@Kilian I think this fixes #820 

Once we land this, we can officially recommend it as a stable alternative to `https://atom.io/download/electron/index.json`.

## Doc Histories

Every doc now has a page that lists all the versions of Electron that include it. Just add `/history` to any docs page, e.g. https://electron-website-pr-960.herokuapp.com/docs/api/browser-window/history

Here's what it looks like:

<a href="https://electron-website-pr-960.herokuapp.com/docs/api/browser-window/history">
screen shot 2017-12-15 at 12 57 16 pm
</a>

## Docs Footer

Before:

screen shot 2017-12-15 at 1 08 12 pm

After:

screen shot 2017-12-15 at 1 06 36 pm


zeke:
Tests should be passing now. This is ready for 👀

script/releases
```diff
@@ -14,24 +16,40 @@ github.authenticate({
...
+
+async function getReleases () {
 
-  return Promise.all(res.data
+  // determine number of pages
+  const countRes = await github.repos.getReleases(
+    Object.assign(ghOpts({per_page: 1}))
+  )
+  const count = Number(parseLinkHeader(countRes.meta.link).last.page)
+  const pagesToFetch = Math.ceil(count / 100)
```
vanessayuenn: Might want to add a safety here so we won't be dividing `NaN` by 100.

script/releases
```diff
@@ -14,24 +16,40 @@ github.authenticate({
...
+  
+  // paginate
+  let releases = []
+  for (let i = 1; i <= pagesToFetch; i++) {
+    const batch = await await github.repos.getReleases(ghOpts({page: i}))
```
vanessayuenn: `await await`?

zeke: Ha! Super-patient. Miraculously, that worked. Fixed now.

script/releases
zeke: 👍 fixed

middleware/browsersync.js
```diff
@@ -13,7 +13,8 @@ module.exports = function browsersync () {
       path.join(__dirname, '../data/**/*'),
       path.join(__dirname, '../js/**/*')
     ],
-    logSnippet: false
+    logSnippet: false,
+    logLevel: 'silent'
```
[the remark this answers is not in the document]
zeke: I just got sick of seeing this every time the server starts:

screen shot 2017-12-15 at 2 26 51 pm

But you're right. It's probably better that everyone knows about browsersync when they start the server. I've re-enabled logging.

zeke:
@vanessayuenn what do you think about the `/docs/api/foo?history` URL style?

Would it be better as `/docs/api/foo/history` ?

vanessayuenn:
@zeke I was _just_ thinking that! Yea I think `/docs/api/boo/history` makes more sense since it's a completely different view from `/docs/api/foo`.

zeke:
Okay I changed the URL structure to `/history` instead of `?history`, and updated the original post above accordingly.

Example: https://electron-website-pr-960.herokuapp.com/docs/api/touch-bar/history

MarshallOfSound:
> https://electron-website-pr-960.herokuapp.com/docs/api/touch-bar/history

This is interesting, the page says the API was introduced in `1.6.3` but only has docs links for `1.6.5` and onwards? 😕 Is there something going wrong there?

zeke:
@MarshallOfSound good eye. I'll investigate.

Kilian:
Electron release data:

* the size is not a problem, for me at least. I do have a cronjob running that fetches this file every day though, so that's ~170MB a month that file will take in bandwidth from me alone. For electron-to-chromium I really only need the depencencyVersions bit, which is much smaller. To save bandwidth, maybe that can be split off separately?
* This file doesn't seem to contain any of the beta versions. The old file does, for the most recent 1.8.2-beta.3 version at least. Is that a conscious choice? Using the beta designation we can support the 'unreleased' tag in browserslist.

zeke:
@Kilian I've updated this PR to use the new `electron-releases` module I just wrote. Find out all about that at https://github.com/electron/electron-releases#readme

I've added a feature to the `/releases.json` endpoint that lets you pick the specific props you want from each object using the comma-delimited `pick` query param:

- [/releases.json](https://electron-website-pr-960.herokuapp.com/releases.json) gives you everything.
- [/releases.json?pick=tag_name,released_at,dependencyVersions.node](https://electron-website-pr-960.herokuapp.com/releases.json?pick=tag_name,published_at,dependencyVersions.node) gives you a lighter feed. Notice the `.` dot which you can use to pluck deep properties if needed.

What do you think?

Kilian:
That pretty nice, and now also included the beta versions, perfect.

There's something iffy though, https://electron-website-pr-960.herokuapp.com/releases.json?pick=dependencyVersions.chrome,dependencyVersions.version has a couple of empty objects and looking at the full file, those are the previous beta versions of 1.8.2, which are missing the dependencyVersions key. Is that on purpose?

Kilian:
@zeke probably cleaner that way indeed. As of https://github.com/Kilian/electron-to-chromium/releases/tag/v1.3.30 I now use electron-releases :)

zeke:
@Kilian 👍

PR #974, opened 2017-12-21
handle docs that have no history

### LINKED ISSUE #973: 500 error on docs history page
/docs/development/upgrading-node/history is causing a 500

```
TypeError: Cannot read property 'replace' of undefined
    at doc.releases.req.context.releases.filter.release (/Users/z/git/electron/electronjs.org/routes/docs/history.js:13:52)
```

zeke:
Fixes #973


vanessayuenn:
Nice 🎉

PR #975, opened 2017-12-21
Translate doc

### LINKED ISSUE #867: Add a "Translate this Doc" link to docs footer

pierreneter:
Close #867 
Related to #960, #910


zeke:
@pierre I would prefer that this is handled in the electron-i18n repo and exported as a new property on the module, which you should now be able to work on without any secrets like `CROWDIN_KEY`. Does that make sense?

zeke:
In here somewhere: https://github.com/electron/electron-i18n/blob/205a0e1be0506aacf01650df3523199e92907066/script/build-module.js#L40

package.json
```diff
@@ -36,7 +36,7 @@
     "dotenv": "^4.0.0",
     "electron-api-historian": "1.1.0",
     "electron-apps": "1.2214.0",
-    "electron-i18n": "0.61.0",
+    "electron-i18n": "github:electron/electron-i18n#translate-doc-url",
```
pierreneter: Related to: https://github.com/electron/electron-i18n/pull/218

zeke:
I shipped https://github.com/electron/electron-i18n/pull/218/ and bumped the i18n version here.

zeke:
I pushed up some fixes. This should be good to go!

Let's wait for the Heroku deployment and give it a go before we ship it.

zeke:
This is great! If anyone wants to try it out:

1. Go to https://electron-website-pr-975.herokuapp.com/languages
1. Pick a language
1. Visit a doc, e.g. https://electron-website-pr-975.herokuapp.com/docs/api/browser-window
1. Scroll to the bottom
1. Click "Translate this doc" (or whatever this string is in your language)

You'll be taken directly to the right page on Crowdin! 🎉

zeke:
@pierreneter is this ready to go? If so, please remove the WIP label so all checks will pass.

vanessayuenn:
Hrmm 🤔 the links on the Chinese docs (and maybe other languages) are not working -- they're pointing to the wrong locale.

For example, https://electron-website-pr-975.herokuapp.com/docs/api/power-monitor, the "Translate this doc" link is pointing to `https://crowdin.com/translate/electron/92/en-zh-CN`.

pierreneter:
Oooops. Sorry. I forgot

pierreneter:
Fix it soob

zeke:
Replaced by https://github.com/electron/electronjs.org/pull/1158 which is assembling URLs using the new [Crowdin File ID property](https://github.com/electron/electron-i18n/pull/230). Thanks, @pierreneter!

PR #999, opened 2018-01-06
New app layout

### LINKED ISSUE #995: Improve layout of app pages
@donokuda made a great start at app-store-ifying the app pages in #822, but time has passed and much has changed since that WIP kicked off, so I've closed it. Let's use some of the ideas in that PR and take another shot at them:

- move more useful content "above the fold" on desktop. See https://user-images.githubusercontent.com/1174461/32741877-8bf0520a-c85c-11e7-8ee9-75a38f6ef160.png
- display the README by default
- what else?

MarshallOfSound:
Closes #995

Of note here

* README always visible
* App metadata (the left column) is always visible if the screen supports it, that bar becomes sticky as you scroll down the readme
* Works on mobile screens
* Tweaked styles across the board
* Better use of full screen width 👍


MarshallOfSound:
Example: https://electron-website-pr-999.herokuapp.com/apps/google-play-music-desktop-player

scripts/update-app-download-links.js
```diff
@@ -26,6 +28,6 @@ module.exports = function updateAppDownloadLinks () {
   }
 
   if (links.length > 1) {
-    document.querySelector('#view-all-downloads').style.display = 'inline-block'
+    document.querySelector('#view-all-downloads').style.display = 'block'
```
vanessayuenn: `Download` section would be empty if there's only 1 available distribution that is not matching current OS.

zeke: The download section is not displayed unless there are downloads, because there's a conditional in the handlebars template: https://github.com/electron/electronjs.org/blob/cff9f468999d018d0505ee1dbf14b6155df17635/views/apps/show.html#L92-L93

vanessayuenn: That's what I expected to happen, but I don't think this is the case? (see https://electronjs.org/apps/protegopdf for an example)

zeke: Doh! Opened https://github.com/electron/electronjs.org/issues/1221

Issue #1022, opened 2018-01-18
404s


vanessayuenn:
This issue captures all the links that lead to nowhere (404) on the electronjs. When we finally clear out this list, we can look into incorporating the work done in #1023 into tests.

```
/OpenNewsLabs/autoEdit_2/blob/1.0.7/www.autoEdit.io (referrer: /apps/autoedit)
/history (referrer: /docs/tutorial)
/docs/tutorial/unity-launcher (referrer: /docs/tutorial/desktop-environment-integration)
/docs/README (referrer: /docs/tutorial/quick-start)
/docs/api/web-view (referrer: /docs/tutorial/security)
/docs/api/optional (referrer: /docs/api/app)
/docs/api/web-view-tag (referrer: /docs/api/web-contents)
/docs/api/structures (referrer: /docs/api/structures/crash-report)
/docs/api/structures/optional (referrer: /docs/api/structures/jump-list-category)
```

sameoldmadness:
The first two links on [Developing Electron](https://www.electronjs.org/docs/development/README) page also lead to 404.

- Code of Conduct (https://www.electronjs.org/CODE_OF_CONDUCT)
- Contributing to Electron -> (https://www.electronjs.org/CONTRIBUTING)

erickzhao:
@sameoldmadness Thanks for the heads-up!

DCzajkowski:
Does this issue cover missing `/api/*/history` pages in the new docs or should I create a new issue?

erickzhao:
@DCzajkowski The API history feature hasn't been implemented in the beta docs yet. Please file a new issue in the tracker here https://github.com/electron/electronjs.org-new!

PR #1023, opened 2018-01-18
add links checkers the crawls the site for 404s

vanessayuenn:
Results compile to #1022. Right now the command needs to be run manually, but after we've fixed the known 404s, we can look into adding this as the last step of the build process.


package.json
```diff
@@ -14,7 +14,8 @@
     "release": "node script/release",
     "dev": "cross-env NODE_PATH=. nodemon -e js,scss,yml,md server.js",
     "prepack": "check-for-leaks",
-    "prepush": "check-for-leaks"
+    "prepush": "check-for-leaks",
+    "linkschecker": "NODE_PATH=. NODE_ENV=test node scripts/links-checker.js"
```
zeke: Why do `NODE_PATH` and `NODE_ENV` need to be set?

scripts/links-checker.js
```diff
@@ -0,0 +1,32 @@
...
+const app = require('../server.js')
+const port = Number(process.env.PORT) || argv.p || argv.port || 5000
+const host = process.env.HOST || `http://localhost:${port}`
+
+const crawler = require('simplecrawler')(host)
+crawler.parseScriptTags = false
+
+crawler.addFetchCondition((queueItem, referrerQueueItem, cb) => {
+  cb(null, !queueItem.path.match(/\/images|styles\//gi))
+  // cb(null, queueItem.path.match(/\/docs\//gi))
```
zeke: This comment can be removed.

scripts/links-checker.js
```diff
@@ -0,0 +1,32 @@
...
+  const $ = cheerio.load(buffer.toString('utf8'))
+  return $('a[href]').get().map((elem) => $(elem).attr('href'))
+}
+
+crawler.on('fetch404', (queueItem, response) => {
+  console.warn(`ERROR(404) fetching ${queueItem.path} (referrer: ${queueItem.referrer.replace(host, '')})`)
+})
+
+crawler.on('complete', () => {
+  console.log('fin')
```
zeke: 👌

package.json
vanessayuenn: it's so the server would start properly, and without the browsersync stuff.

PR #1031, opened 2018-01-22
Support a new `language` query param

zeke:
This PR allows a `?language=` query param to be added to any URL to view that URL in the target language. This does not have an effect on the visitor's language settings (or lack thereof). In other words, if my language is English and I view `/docs/api/app?language=fr-FR`, I will only see that one page in French. If I click any links or remove the query param from the URL, content will still be displayed in English.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -267,10 +267,12 @@ describe('electronjs.org', () => {
     })
   })
 
-  test('/languages', async () => {
-    const $ = await get('/languages')
-    $('h1').text().should.eq('Languages')
-    $('body').text().should.include('global developer community')
+  describe('languages', () => {
+    test('/languages', async () => {
+      const $ = await get('/languages')
+      $('h1').text().should.eq('Languages')
+      $('body').text().should.include('global developer community')
+    })
   })
 
   test('redirects for date-style blog URLs', async () => {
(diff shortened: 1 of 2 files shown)


zeke:
Example URL: https://electron-website-pr-1031.herokuapp.com/docs/api/browser-window?language=zh-CN

zeke:
This should make it easier for us to diagnose issues during our upcoming [call with the Crowdin team](https://github.com/electron/electron-i18n/issues/215)

test/index.js
```diff
@@ -267,10 +267,20 @@ describe('electronjs.org', () => {
...
+      $('h1').text().should.eq('Languages')
+      $('body').text().should.include('global developer community')
+    })
+
+    test('language query param for one-off viewing in other languages', async () => {
+      let $ = await get('/docs/api/browser-window')
+      $('body').text().should.not.include('fenêtres')
+
+      $ = await get('/docs/api/browser-window?language=fr-FR')
+      $('body').text().should.include('fenêtres')
```
vanessayuenn: Should we swap the order of these two tests, so we know the `?language=fr-FR` param isn't affecting the language stored in the cookies?

zeke: Good call. Done.

zeke:
Oops. The query param is persisting in the cookie. Turns out this is a feature of `express-request-language` that I hadn't noticed: https://github.com/tinganho/express-request-language/blob/master/README.md#queryname-optional-object

I guess that shows that this testing strategy doesn't account for user state.

Working on a fix.

PR #1033, opened 2018-01-23
excerptize protocol handler post intro

zeke:
This is a cosmetic change that adds an `---` (HR tag) to prevent this entire post from being displayed on the blog index.


zeke:
Looks like this now:

screen shot 2018-01-22 at 7 51 39 pm

PR #1034, opened 2018-01-23
add lang query param and document it in contributing

zeke:
This PR is a followup to https://github.com/electron/electronjs.org/pull/1031 that changes the one-off param from `language` to `lang` to avoid stepping on the toes of the existing default `language` query param name that [`express-request-language` allows](https://github.com/tinganho/express-request-language/blob/master/README.md#queryname-optional-object). This also adds docs to CONTRIBUTING.md so folks know they can assemble shareable translated URLs.


zeke:
For some reason the use of `supertest-session` is causing the tests to hang, even though they're passing.

zeke:
> We should consider incorporating that into the UI

Yeah I was thinking that too. "View this page in..."

zeke:
@rjs was kind enough to publish a [fix](https://github.com/rjz/supertest-session/issues/28) to supertest-session. This should go green now.

vanessayuenn:
@zeke the `package-lock.json` was causing `npm i` to fail at updating `supertest-session` for some reason. It looks ok now, so merging this PR. :)

PR #1037, opened 2018-01-23
Link to search from nav bar

vanessayuenn:
This will make the `/search` route accessible to public! 🎉  Also, I improved the description a bit to be more explicit about which data source the users are searching within.

### DIFF
--- a/data/locale.yml
+++ b/data/locale.yml
@@ -198,4 +198,9 @@ _404:
   please_file_an_issue: please file an issue.
 
 search:
-  description: Search within Electron documentations and community.
+  description:
+    start: Find Electron resources from
+    docs: <a href="/search/docs">documentations</a>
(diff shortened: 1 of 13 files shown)


zeke:
Hey I took the liberty of making some stylistic tweaks:

- turned headings into links
- added octicons to headings
- made results more compact, with one line of text per result
- lowered opacity for stuff that doesn't need to be loud, like result counts

Here's what it looks like now:

screen shot 2018-01-23 at 8 24 46 pm

zeke:
Passing the review baton back to you, @vanessayuenn 🤝

pierreneter:
`span.label` influences the `label` of the `button` at the homepage. Just remove `class="label"` at home to fix it. I will push a commit.

snap1

vanessayuenn:
Niceeeee! That looks great @zeke. 💯 
image

PR #1043, opened 2018-01-25
Upgrade primer breadcrumbs

jonrohan:
There were a few things I noticed and fixed.

When upgrading `primer-breadcrumbs` before. The build was failing because you were importing `primer-breadcrumbs/lib/breadcrumbs.scss` directly. Importing `primer-breadcrumbs/index.scss` is recommended because it includes the required `primer-support` file.

When changing that, the sass middleware didn't know where to find it. So adding `node_modules` to the `includePaths` for the sass middleware tells it where to look when `@import "primer-support/index.scss"` is in the css.

Adding the `includePaths` removed the need to have `@import "../node_modules/primer-breadcrumbs/...` in the `styles/index.scs` file. So I updated that also.

The rest of the `primer-css` modules are so out of date, I would recommend updating them in the same way. I could easily do that in this pr, but there are potential visual breaking that could occur on the site which would take more time to debug. I would recommend starting by replacing any bower modules with node modules that haven't changed much from what you have.

closes https://github.com/primer/primer/issues/441

cc @zeke @broccolini @vanessayuenn

### DIFF
--- a/middleware/sass.js
+++ b/middleware/sass.js
@@ -6,6 +6,7 @@ module.exports = function () {
     src: path.join(__dirname, '../styles'),
     response: true,
     debug: true,
+    includePaths: path.join(__dirname, '../node_modules'),
     // outputStyle: 'compressed',
     prefix: '/styles'
   })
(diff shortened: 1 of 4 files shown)


zeke:
Thanks for digging into this @jonrohan. I will follow up with a test to help prevent future regressions.

PR #1044, opened 2018-01-25
test that the stylesheet responds with 200

zeke:
This is a followup to #1043 that adds a test to verify that `/styles/index.css` responds with a 200 HTTP status code. This should help prevent us from shipping stylesheet regressions in the future.

@jonrohan I see this warning show up several times in the test output:

```
WARNING: Unfortunately, no value could be retrieved from `xs`. Please make sure it is defined in `$breakpoints` map.
Backtrace:
	node_modules/primer-support/lib/mixins/layout.scss:17, in mixin `breakpoint`
	bower_components/basecoat/scss/_utility.scss:27
```

How would recommend solving it?

cc @broccolini

### DIFF
--- a/bower_components/basecoat/scss/_variables.scss
+++ b/bower_components/basecoat/scss/_variables.scss
@@ -20,6 +20,19 @@ $grid-breakpoints: (
   xl: 1200px
 ) !default;
 
+// HACK! $grid-breakpoints and $breakpoints are dupes. See https://git.io/vNPI4
+$breakpoints: (
+  // Extra small screen / phone
+  xs: 0,
+  // Small screen / phone
+  sm: 544px,
+  // Medium screen / tablet
+  md: 768px,
+  // Large screen / desktop
+  lg: 992px,
+  // Extra large screen / wide desktop
+  xl: 1200px
+) !default;
 
 // Grid containers
 //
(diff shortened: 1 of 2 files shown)


jonrohan:
@zeke unfortunately some conflicting variables/mixins between basecoat and primer. The harder better solution would be to drop some of those basecoat files for the same ones in new primer. The quick easy fix would be to duplicate `$grid-breakpoints` in `bower_components/basecoat/scss/_variables.scss` and give it `$breakpoints` name.

```
❯ git diff
diff --git a/bower_components/basecoat/scss/_variables.scss b/bower_components/basecoat/scss/_variables.scss
index 19d06893..8e0e1a14 100644
--- a/bower_components/basecoat/scss/_variables.scss
+++ b/bower_components/basecoat/scss/_variables.scss
@@ -20,6 +20,19 @@ $grid-breakpoints: (
   xl: 1200px
 ) !default;
 
+$breakpoints: (
+  // Extra small screen / phone
+  xs: 0,
+  // Small screen / phone
+  sm: 544px,
+  // Medium screen / tablet
+  md: 768px,
+  // Large screen / desktop
+  lg: 992px,
+  // Extra large screen / wide desktop
+  xl: 1200px
+) !default;
+
```

zeke:
Thanks. Went with the "quick easy fix" 🙈

jonrohan:
I thought so 😏

PR #1048, opened 2018-01-26
update electron-releases

vanessayuenn:
Looks like because the releases came out so soon after one another during the security patching, the earlier version of `electron-releases` got merged instead of the latest in #1030, resulting in outdated info on https://www.electronjs.org/releases.


zeke:
Thanks @vanessayuenn

PR #1053, opened 2018-01-30
Language toggle for doc sections

### LINKED ISSUE #886: Show original strings from translated docs
While documentation i18n is a great way to expand the reach of Electron, it's a long-term effort, and as a matter of fact, it is hard to completely avoid improper or outdated translations. As a result, users that read docs in another language often open up the translated docs and the English docs side by side for comparison. 

The experience of reading translated docs will be improved tremendously if the original string of the docs can be easily accessible from within the translated docs. 

Prior Art:
[Angular Chinese Docs](https://angular.cn/)

vanessayuenn:
This implements the idea in #886 to make the original documentation string of the docs more easily accessible from within the translated docs. It allows for users to toggle _each doc section_ between their selected language, and the original English version. While some prior art has enabled the original-translation-toggle _per sentence_, it's very hard to do so for the electron docs, because of how the documentation and its translations are structured. 

All feedback is, of course, welcome, but critiques on the ux would be especially helpful. 

kapture 2018-01-30 at 17 40 27

~~Note: this PR is dependent on https://github.com/electron/electron-i18n/pull/222, so PR deployment is not possible right now.~~

For example: https://electron-website-pr-1053.herokuapp.com/docs/tutorial/updates?lang=zh-CN

Close #886


zeke:
The Heroku build is failing because:

```
npm ERR! Could not install from "../electron-i18n" as it does not contain a package.json file.
```

What about installing from a GitHub branch instead, like `electron/electron-i18n#foo` ?

vanessayuenn:
@zeke Hrmm looks like that's not possible because of how `electron-i18n` is released? The branch won't have `index.json` since it's git-ignored. 🤔

vanessayuenn:
Yay it's working with the latest `electron-i18n` release!  🎉🍻🍾

zeke:
🎉

PR #1068, opened 2018-02-02
fix app gallery icons

### LINKED ISSUE #1067: Images not loading on Homepage
Just change [this line](https://github.com/electron/electronjs.org/blob/master/views/apps/index.html#L43) from `data-src` to `src` & it'll work.

ckerr:
fix suggested by @deadcoder0904

Closes #1067


pierreneter:
Hey, I think it's ok with `data-src` because of we are using [`lazy-load-images`](https://github.com/electron/electronjs.org/blob/master/scripts/lazy-load-images.js). There is something made it stop work. I will check.

pierreneter:
Because there was an issue at `update-demo-app-download-link.js`, `lazy-load` script was not running.

vanessayuenn:
Hey @ckerr, yea this will break lazy-load for the images in /apps. Closing as #1069 should have the original issue resolved.

PR #1072, opened 2018-02-05
bump electron-releases

zeke:
Fixes https://github.com/electron/electronjs.org/issues/1070


zeke:
> Does it have to be exactly 2.8.0

No it doesn't. I just set it that way knowing that greenkeeper only uses specific version numbers and will eventually blow that range away.

zeke:
Not seeing an automated Heroku deployment on this. The bots are failing us!

PR #1078, opened 2018-02-07
fix(package): update electron-i18n to version 0.77.0

zeke:
Closes #1074


zeke:
See https://github.com/electron/electronjs.org/pull/1079

PR #1094, opened 2018-02-13
General search errors handling

vanessayuenn:
Add handling for other graphql errors.


zeke:
Nice! Thanks for taking the initiative and shipping this quickly 🚀

Issue #1107, opened 2018-02-15
Make `lang` param more permissive


zeke:
- https://electronjs.org/docs/tutorial/security?lang=en-us does not work.
- https://electronjs.org/docs/tutorial/security?lang=en-US does work.
- https://electronjs.org/docs/tutorial/security?lang=en _should_ work.

cc @groundwater

FranzDeCopenhague:
I can help on that

FranzDeCopenhague:
What supposes to happen with a none translated region? For example Spanish from Argentina https://electronjs.org/docs/tutorial/security?lang=es-AR
My guess is redirecting to lang=es-ES

zeke:
I would start by looking at how https://github.com/tinganho/express-request-language parses the accept-language header. As I recall it's pretty permissive, by which I mean you can give it a short string like `en` or a locale that is not supported, and it will pick the closest match. Maybe we can re-use whatever it's using.

FranzDeCopenhague:
See above commit, I did apply a different approach. Based on the current Electron locales set, filter the lang out and redirect the request with the correct lang value.

* https://electronjs.org/docs/tutorial/security?lang=en-us redirect to https://electronjs.org/docs/tutorial/security?lang=en-US
* https://electronjs.org/docs/tutorial/security?lang=es-AR redirect to https://electronjs.org/docs/tutorial/security?lang=es-ES

FranzDeCopenhague:
Verified that now works
* https://electronjs.org/docs/tutorial/security?lang=en-us
* https://electronjs.org/docs/tutorial/security?lang=en

zeke:
🎉 Nice work, @FranzDeCopenhague

50Wliu:
I believe this broke Traditional Chinese.  `lang=zh-TW` now redirects to `lang=zh-CN`.

FranzDeCopenhague:
I have fixed this issue with a new pull https://github.com/electron/electronjs.org/pull/1115

FranzDeCopenhague:
@50Wliu @zeke Verified in production now https://electronjs.org/docs/tutorial/security?lang=zh-TW does not redirect

PR #1112, opened 2018-02-16
redirect the request with the correct lang #1107

FranzDeCopenhague:
@zeke, Would you have a look at this new middleware to resolve the lang?

### DIFF
--- a/middleware/lang-resolver.js
+++ b/middleware/lang-resolver.js
@@ -1,0 +1,53 @@
+const i18n = require('electron-i18n')
+const url = require('url')
+var parseUrl = require('parseurl')
+
+const locales = Object.keys(i18n.locales)
+
+var lowerLangRegion = {}
+var lowerLang = {}
+// Keep all the language/region and language in a object
+// with property names in lowercases to resolve the lang code
+locales.forEach(locale => {
+    const s = locale.split('-')
+    const lang = s.shift().toLowerCase()
+    const region = s.join(' ').toUpperCase()
+
+    lowerLangRegion[`${lang}-${region}`] = locale
+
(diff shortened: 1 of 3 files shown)


FranzDeCopenhague:
@zeke I have added the tests.

middleware/lang-resolver.js
```diff
@@ -0,0 +1,56 @@
+const i18n = require('electron-i18n')
+const url = require('url')
+var parseUrl = require('parseurl')
```
zeke: You should be able to use Node's built-in `require('url').parse` here instead of adding a new dependency.

FranzDeCopenhague: I will try. Thank you for the feedback

FranzDeCopenhague:
@zeke parseurl dependency was removed. This pull is ready to go ....

middleware/lang-resolver.js
```diff
@@ -45,7 +48,7 @@ module.exports = function langResolver (req, res, next) {
 
     // redirect only when the resolved lang exists and it is different
     if (rlang && rlang !== req.query.lang) {
-      const parsedUrl = parseUrl(req)
+      const parsedUrl = parseurl(req)
```
50Wliu: Why not `url.parse(req.url)` directly?

FranzDeCopenhague: well, I like parseurl (https://www.npmjs.com/package/parseurl) because it reuses the req._parsedUrl and it has better performance. But @zeke recommended me to avoid adding extra dependencies. So, I did remove it from my original pull but keeping parseurl(req) in case of we use it in the future.

PR #1127, opened 2018-02-21
fix all the 404s for `/docs/api/structure/**/history` routes

vanessayuenn:
This will reduce the 404 counts in #1022 down to 10 😄


zeke:
💯

PR #1129, opened 2018-02-21
Two versions on homepage

### LINKED ISSUE #934: Advertise electron beta version along side stable version
We should show both the stable version and beta version. 

I'm thinking we can do that by adding another line on the front page (see below) where we currently show the stable version number, along with the chromium and node version used in stable, but I'm also open to another approach. 
image

zeke:
Resolves #934 

This is an update to display the `latest` and `beta` version info on the homepage, instead of just `latest`.

Here's a shot at the visual:

screen shot 2018-02-21 at 2 38 59 pm

I was also trying out using devicons for Node.js and Chromium (forgive the redundant info in this transitional screenshot):

screen shot 2018-02-21 at 2 39 54 pm

To Do

- [x] Figure out out how to get each side to fill up 50% of available viewport width
- [ ] Wait for an updated version of [electron-releases](https://github.com/electron/electron-releases) that includes Chrome and Node version data from S3.
- [x] Explore other ways of displaying this info.

@simurai feel free to take a turn on this PR if you have the time and energy!

### DIFF
--- a/data/locale.yml
+++ b/data/locale.yml
@@ -50,6 +50,8 @@ community:
     To add a meetup, edit <code>data/meetups.json</code> and
     <a href="https://github.com/electron/electronjs.org/tree/master/data/meetups.json">make a pull request</a>.
 
+releases: Releases
+
 electron_is_easy:
   title: It's easier than you think
   description:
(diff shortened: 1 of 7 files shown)


zeke:
> What about highlight the difference.

I like that! Do you know what elements / classes to use to get the divs to auto-fill the horizontal space?

simurai:
> Do you know what elements / classes to use to get the divs to auto-fill the horizontal space?

Yeah, now there should be two columns in a wide window, and full-width when narrow:

releases

PR #1134, opened 2018-02-22
Remove scrollbar on release version info box

vanessayuenn:
There's currently an unsightly scrollbar on the release version boxes :( 
(on Chrome and Safari)
image


zeke:
grazie!

PR #1158, opened 2018-02-28
Link docs to Crowdin

### LINKED ISSUE #867: Add a "Translate this Doc" link to docs footer

zeke:
This is a replacement PR for https://github.com/electron/electronjs.org/pull/975 that fixes #867 

Currently failing because:

```
  1) electronjs.org
       docs
         docs footer
           includes a link to translate the doc on Crowdin:

      AssertionError: expected 'https://crowdin.com/translate/electron/63/en-zh-CN' to equal 'https://crowdin.com/translate/electron/63/en-zhcn'
      + expected - actual

      -https://crowdin.com/translate/electron/63/en-zh-CN
      +https://crowdin.com/translate/electron/63/en-zhcn
```

Gonna reach out to Crowdin folks to figure out where we can get the `zhcn` formatted code... we don't currently have that in our i18n locale data. The `stats` object below is what we get from Crowdin's API:

```
> i18n.locales['zh-CN']
{ locale: 'zh-CN',
  languageCode: 'zh-CN',
  languageName: 'Chinese Simplified',
  languageNativeName: '中文',
  countryCode: 'CN',
  countryName: 'China',
  stats: 
   { name: 'Chinese Simplified',
     code: 'zh-CN',
     phrases: '8847',
     translated: '6060',
     approved: '94',
     words: '79936',
     words_translated: '57011',
     words_approved: '520',
     translated_progress: 71,
     approved_progress: 0,
     qa_issues: '157' } }
```

Maybe the answer is just `locale.code.toLowerCase().replace('-', '')` but I'm not sure.


Andrulko:
URLs in language codes that lead to the editor (e.g. `en-zhcn`) are indeed a bit different from the language codes listed here: https://support.crowdin.com/api/language-codes/

They're written with small letters always and there is no hyphen between language and country code in locae codes (e.g. `ptbr` instead of expected `pt-BR`). It's indeed a bit strange and I think it's due to some historical reasons and of course we cannot change this now easily because URLs copied by others earlier won't work (but probably we can do some redirect, at least temporarily).

You've specified some script that should do the following I think: modify all letters to small and modify language codes obtained from https://support.crowdin.com/api/info/ to remove hyphen. It's a good workaround.

Currently there is no API endpoint that can provide you codes we use for editor URLs. The only major difference here is that language dialects doesn't contain `-` char...

Can we please leave this question open for a bit?There is a chance that language codes used in URLs will be updated and missing hyphen will return to its place. Then language codes will totally correspond to those you receive over API and custom scripts on your end won't be required

vanessayuenn:
Hey @Andrulko @zeke I don't have all the context here, but if I remember correctly, `es-ES` was the exception to that rule and linking to editor to translate Spanish documents uses `en-es`.

zeke:
>  linking to editor to translate Spanish documents uses en-es.

Hmm. So maybe if the first part of the locale string matches the last part, then it's reduced to a single code? e.g. `foo-FOO` becomes `foo`?

pierreneter:
> Maybe the answer is just locale.code.toLowerCase().replace('-', '') but I'm not sure.

Yah, that's really not sure, an old way is a case I did at https://github.com/electron/i18n/pull/230 is: `en-${locales[file.locale].stats.code.replace('-', '')` it work fine for almost language cases, expect `es-ES` still return: `en-eses`, it just an exception. `en-es` for true.

pierreneter:
This seems like the code returned from stats is not in sync with the locale code at Crowdin's URL.

zeke:
@Andrulko can you comment on the `es` thing mentioned above? Do all languages that have the same prefix and suffix like `foo-FOO` get condensed into `foo`?

Andrulko:
@zeke Please expect updates on Monday, trying to invent something useful here

zeke:
@Andrulko any updates on this?

Andrulko:
Hi @zeke, it's more a historical reason why language codes in the edtior don't fully match with [Crowdin Language Code](https://support.crowdin.com/api/language-codes/). We cannot change it right now, since bookmarks people may have in their browsers will become broken.

We have extracted language codes which are used in the editor, hope they will help you to match languages accordingly!
[Crowdin Editor Language Codes](https://docs.google.com/spreadsheets/d/15BQy_gmPFUaPoXD5FYkGgUUVYuHGI5SkmYSUyw_u4Dg/edit?usp=sharing)

zeke:
I put those editor language codes into a module: https://github.com/crowdin-modules/crowdin-editor-language-codes

zeke:
Replaced by https://github.com/electron/electronjs.org/pull/1219

PR #1192, opened 2018-03-07
remove github dependencies

vanessayuenn:
We don't appear to actually use this package anywhere in the code. Closes #1191.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,14 +69,6 @@
       "resolved": "https://registry.npmjs.org/after/-/after-0.8.1.tgz",
       "integrity": "sha1-q11PuIP1loFtNRX495HAr0ht1ic="
     },
-    "agent-base": {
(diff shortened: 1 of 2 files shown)


zeke:
Looks like there are others too:

```
~/git/electron/electronjs.org master 7s
$ npx depcheck .
Unused dependencies
* babel-preset-env
* clean-deep
* dotenv
* electron-userland-reports
* express-graphql
* github
* got
* href-type
* mkdirp
* parse-link-header
* primer-breadcrumb
* ua-parser-js
Unused devDependencies
* husky
Missing dependencies
* lib
```

Most of these probably unused, but`depcheck` is not 100% accurate, so we should be careful. For example, `husky` is in fact used by `check-for-leaks`, and `primer-breadcrumb` is used just for its stylesheets.

PR #1204, opened 2018-03-09
Add script to update certain dependencies

### LINKED ISSUE #1046: Should we use greenkeeper-keeper?
Greenkeeper is working well for us, but we have a handful of modules that update very often, namely `electron-i18n` and `electron-apps`. I'm thinking maybe we should set something up to ship Greenkeeper PRs automatically if they pass all checks, like [greenkeeper-keeper](https://github.com/greenkeeper-keeper/reference-instance).

## Pros

- Changes are shipped quickly
- Less monotonous work for humans
- Decreased likelihood of git conflicts in package.json

## Cons

- The site could break! We've seen this a few times with `primer`-related updates that we didn't have tests for.
- We'd have to maintain a new self-hosted service (unless we find a Probot or some other hosted service that does it for us)

What do folks think? Should we try it?

vanessayuenn:
The script added here should be run as a scheduled task periodically on Heroku. It checks for then applies updates for certain dependencies: `electron-i18n` `electron-apps` `electron-api-historian` `electron-releases`. These modules update very often, and are owned by electron org so it should be relatively low risk to auto-update them.

The goal is to lessen the amount of monotonous manual work, and make changes to these packages show up on the website faster. This should resolve #1046.


zeke:
This is great! Looking forward to less noise.

I'm late to the game here, but I think we might run into issues if we're running this as a separate scheduled task in Heroku.. if the two tasks overlap in execution, one will not be able to 'git commit' its changes if there are conflicts. The easiest way to avoid this would be to consolidate this new stuff into the existing `update-source-content.sh` script, and maybe rename that script as it becomes more multi-purpose.

zeke:
I think it would just be a matter of copying over these two lines:

```sh
TRUSTED_DEPENDENCIES="electron-i18n electron-apps electron-api-historian electron-releases"
```

and later

```sh
npm update $TRUSTED_DEPENDENCIES
```

vanessayuenn:
@zeke 🤔will that issue happen even when the two scripts are run on different heroku apps with different repos?

zeke:
oh derrr! That's what I get for dipping into PR review when I'm on vacation. Yep. Two different repos. Apologies. Proceed!

PR #1208, opened 2018-03-14
Change emails to @electronjs.org domain

groundwater:
CoC should go to `coc@electronjs.org`
Security should go to `security@electronjs.org`


zeke:
Thanks!

PR #1210, opened 2018-03-20
Add OpenCollective donors page

groundwater:
Created a `/donors` page. The contents is rendered from opencollective.com so it may or may not have a style that matches our theme at first. We can probably improve it down the road.

- [x] `/donors` route
- [x] menu item
- [x] donors localization

See: https://opencollective.com/electron

### DIFF
--- a/data/locale.yml
+++ b/data/locale.yml
@@ -2,6 +2,7 @@ tagline: Build cross platform desktop apps with JavaScript, HTML, and CSS
 
 nav:
   docs: Docs
+  donors: Donors
   blog: Blog
   community: Community
   apps: Apps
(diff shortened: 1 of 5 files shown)


data/locale.yml
```diff
@@ -197,6 +197,9 @@ pages:
 _404:
   page_not_found: Page not found.
 
+donors:
+  donors: Donors
```
zeke: There's a section in locale.yml for page metadata like title, description etc. I think it's called `pages`. Put this there and you won't need to set `req.context.page.title` in the route handler.

groundwater:
- [x] maybe just ditch the package-lock.json changes
- [x] See comment above about locale.yml page data

@vanessayuenn 👍 
@zeke 👍

zeke:
💰

PR #1213, opened 2018-03-22
disable donors in nav

zeke:
Let's not link to the @opencollective donors page until we have donors.

cc @groundwater


zeke:
Yep. @groundwater is going to write a blog post to explain what we're up to. Closing.

PR #1215, opened 2018-03-26
default to English if lang param is invalid

zeke:
Fixes 500 errors for requests like`//index.php?lang=Cn&index=0000`

image

### DIFF
--- a/middleware/context-builder.js
+++ b/middleware/context-builder.js
@@ -17,6 +17,10 @@ module.exports = function contextBuilder (req, res, next) {
     req.language = req.query.lang
   }
 
+  // If request language does not exist, fall back to English
+  // e.g. /?lang=Foo
+  if (!i18n.website[req.language]) req.language = 'en-US'
+
   const localized = i18n.website[req.language]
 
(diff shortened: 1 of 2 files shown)


FranzDeCopenhague:
@Zeke this change may have an impact on https://github.com/electron/electronjs.org/issues/1107
https://github.com/electron/electronjs.org/blob/master/middleware/lang-resolver.js

zeke:
The `langResolver` middleware is before the `contextBuilder`, so it's executed first. We should be good:

https://github.com/electron/electronjs.org/blob/1adcb8c362c16e104a687179536b263a7cca8b1e/server.js#L57-L58

Also you wrote a bunch of tests for language resolution, correct? If they were affected we'd see some test failures.

FranzDeCopenhague:
Yes, you are right, all checks have passed

FranzDeCopenhague:
This is the first time I do merge a pull request. Do I have to click on "Delete branch" button as well?

codebytere:
@FranzDeCopenhague it's not required, but it helps to keep the branch list clean and tidy

FranzDeCopenhague:
@codebytere Thank you!

PR #1219, opened 2018-04-03
Translate on Crowdin

### LINKED ISSUE #867: Add a "Translate this Doc" link to docs footer

zeke:
This is a replacement PR for #1158, the git history of which was flummoxed. Fixes #867

At long last, this PR adds a link to the footer of every doc page that takes the user straight to the right Crowdin editor page for the given file and language.


zeke:
Gah. Not actually ready yet. Not working for Indonesian, Filipino, and probably others.

zeke:
Ok fixed it. Ready for review.

PR #1223, opened 2018-04-14
[WIP] Obfuscate contact emails

codebytere:
Started this as a little spike on what might tone down the flood of spam emails we're currently getting on our contact emails. 

It converts each ascii email address into its equivalent decimal entity set so that it can't be so easily scraped. 

Please let me know thoughts or if there might be a better way to go about this! I may have also missed a location; i'm not sure if this will work in blog posts.

/cc @MarshallOfSound @zeke

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -34,7 +34,7 @@ This Code of Conduct applies both within project spaces and in public spaces whe
 
 ## Enforcement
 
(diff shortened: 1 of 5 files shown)


zeke:
It'd be nice if we could do this in a way where the emails were intact in the source code.

Here's an express middleware that does the obfuscation: https://github.com/awolk/express-email-obfuscate

codebytere:
Closing in favor of #1243

PR #1224, opened 2018-04-15
Update link to Code Signing post

mkaz:
Update the domain for the Code Signing post, redirects are currently in place but will eventually expire.


zeke:
Thanks!

PR #1248, opened 2018-05-09
search improvements

echjordan:
https://electron-website-pr-1248.herokuapp.com


zeke:
Not sure why the Heroku deploy didn't work. I will dig into that.

Heads up: It looks like CI is now failing because of a change in the HTML structure:

```
  1) localized views
       every localized string is used in a view (except for pages.* and _404.*):
     AssertionError: expected [ Array(100) ] to include 'nav.search'
```

zeke:
[status.heroku.com](http://status.heroku.com/) says Heroku has the hiccups. I pushed up an empty commit to trigger another deployment.

MarshallOfSound:
Looks :awesome: so far 😀 Just wanted to drop by and say it would be 💯 if you could prefix searches with things to restrict results.

E.g. `doc: ` to limit searches to just the API documentation or `app: ` to search for apps 🤔 

Normally you would want to search for one or the other so narrowing that down would be cool 😄

scripts/search.js
```diff
@@ -0,0 +1,153 @@
...
+
+  function showHits () {
+    document.getElementById('hits').style.display = 'flex'
+  }
+
+  function hideHits () {
+    document.getElementById('hits').style.display = 'none'
+  }
+
+  function checkIfChild (parentElement, checkElement) {
```
echjordan: @zeke HALP. trying to recursively check for any child of "hits" to fix click functionality. something is broken. Line 135 calls this function.

PR #1252, opened 2018-05-14
Add note to Electron 2.0 release notes about dropping mips64el support

malept:
This should also be added to https://github.com/electron/electron/releases/tag/v2.0.0


zeke:
Thanks 👍

PR #1260, opened 2018-05-22
display master branch commit for tutorials

zeke:
For API docs:

screen shot 2018-05-22 at 11 14 15 am

For Tutorials:

screen shot 2018-05-22 at 11 14 30 am


zeke:
Currently we're displaying API docs from the `latest` version of Electron, and tutorials from the master branch. This was done during a time when we didn't see a new `latest` for several weeks, and wanted to surface tutorial content on the website without having to wait for a new stable release.

Eventually we should move to a model where site users can select a major version of Electron (1.x, 2.x, 3.x, etc) and see all the docs that correspond to that version. See https://github.com/electron/i18n/issues/288#issuecomment-391090702 for progress on that.

PR #1269, opened 2018-06-03
search with your keyboard

zeke:
This PR adds the new [search-with-your-keyboard](https://github.com/electron/search-with-your-keyboard) module.

Key | Action
--- | ------
<kbd>/</kbd> | Focuses the search input.
<kbd>esc</kbd> | Focuses the search input and clears it.
<kbd>down</kbd> | Adds an `active` class to the next result.
<kbd>up</kbd> | Adds an `active` class to the previous result. If already on the first search result, the search input is focused.
<kbd>enter</kbd> | Sets `window.location` to the `href` of the first `<a>` tag in the current `.active` result, if present.

I snuck in a a few other changes too, like style tweaks and a change to the search order.

### DIFF
--- a/styles/_search.scss
+++ b/styles/_search.scss
@@ -65,10 +65,21 @@ h2 .faint {
   margin-left: 20px;
 }
 
+// #search-hint{
+//   position: relative;
+//   right: 7%;
+//   top: 5px;
+// }
+
+// #search-hint.active {
+//   display: none;
+// }
+
 .nav-search:focus {
   background-color: white;
   color: $jumbo-bg-color;
 }
+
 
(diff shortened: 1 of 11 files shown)


zeke:
Thanks for the feedback @HashimotoYT 👍

styles/_search.scss
```diff
@@ -65,11 +65,22 @@ h2 .faint {
   margin-left: 20px;
 }
 
+// #search-hint{
```
echjordan: This is my own proposed interaction for adding a #search-hint much like the one on github, it's messing with CSS currently so commented but I left in for future use.
screen shot 2018-06-05 at 9 46 00 am

Issue #1278, opened 2018-06-06
Add GitHub-style "hint" to search input so users know about keyboard focus


zeke:
![image](https://user-images.githubusercontent.com/2289/41047555-15c53ed6-69ad-11e8-832a-4604fab06c1c.png)

echjordan:
The branch 'search-with-hint' works except that the input jumps a bit on click? Any ideas @zeke

PR #1280, opened 2018-06-07
proposed search hint w github.com asset

echjordan:
Just a little PR for adding a hint in the search nav, like this! 
screen shot 2018-06-07 at 11 39 43 am

The hint disappears on focus of the input. Could use its own nicely colored asset :)


echjordan:
This PR also addresses #1278 and conveniently covers #1277

zeke:
Something about this change is causing the search input to jump horizontally when focus/blurred. Also there are a few conflicts. Otherwise, looks good.

PR #1294, opened 2018-06-28
Fix logo

simurai:
> @arcanemagus This is _super_ minor, I just happened to notice it but one of the 'swooshes" in the Electron logo animation on https://electron.org sticks just a _tiny_ bit into the circle at the end.

This PR moves the bottom right circle a bit to the right so that the connecting line isn't visible anymore.

Before | After
--- | ---
image | image


zeke:
Thanks 💛

PR #1306, opened 2018-08-09
Search placeholder is now translable.

Voknehzyr:
I added ability to translate placeholder in search bar to another languages.


zeke:
Thanks!

PR #1318, opened 2018-08-24
fix: use better contrastsecurity.com URL

ckerr:
As suggested by Matt.


zeke:
Matt === @mgc ?

ckerr:
Good point. I should've @matt- instead of subtweeting him -- I was being lazy and hadn't looked up his GitHub handle :slightly_smiling_face:

PR #1321, opened 2018-09-12
fix: add missing 'pluralize' module

zeke:
Fixes https://github.com/electron/electronjs.org/issues/1320


zeke:
This is fixed now.

Somehow the `pluralize` module got removed from package.json and it blew up the browserify build. 

See https://electron-website-pr-1321.herokuapp.com and https://electron-website-pr-1321.herokuapp.com/scripts/index.js

Issue #1323, opened 2018-09-15
update-electron-dependencies script failing


zeke:
The `npm run update-electron-dependencies` scheduled Heroku task is failing with `Error: spawn mocha ENOENT`. This is causing `electron-releases` to fall behind on the site.

```
+ git commit -am 'chore: update electron-releases'
[master 3813fa2] chore: update electron-releases
 2 files changed, 4 insertions(+), 4 deletions(-)
+ npm test

> electronjs.org@2.0.0 test /app/website
> cross-env NODE_PATH=. NODE_ENV=test mocha --timeout 5000 && standard --fix

events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: spawn mocha ENOENT
```

cc @ckerr @sofianguy

zeke:
Note: The [electron-releases](https://www.npmjs.com/package/electron-releases) module itself is up to date.

zeke:
Looking at the [recent commit history](https://github.com/electron/electronjs.org/commits/master) for this repo, the script was working fine a few days ago:

![image](https://user-images.githubusercontent.com/2289/45582045-0380a800-b85e-11e8-8468-48cea57c50b9.png)

And judging by the papertrail logs, the first occurrence of this error was the 12th (two days ago):

![image](https://user-images.githubusercontent.com/2289/45582227-fc0ece00-b860-11e8-8c5b-b4950425bb97.png)

I suspect this is because PR https://github.com/electron/electronjs.org/pull/1321 added a module to `package.json` but didn't include the subsequent changes to `package-lock.json`, and `npm ci` is quietly failing to install properly, causing `mocha` to not exist and the `cross-env`-wrapped script to fail without a useful error message. PR https://github.com/electron/electronjs.org/pull/1324 may solve this. 🤞

zeke:
@vanessayuenn and I spent some time debugging our update-electron-dependencies script, but couldn’t figure it out. Then we remembered that Dependabot has a feature for auto-merging whitelisted dependencies! So we’ve enabled dependabot on the repo, and configured it to auto-merge its own PRs for electron-i18n, electron-apps, electron-releases, and electron-api-historian. We’ve also disabled Greenkeeper on the website (which actually hasn’t been working for a while anyway, because the GitHub App requested updated permissions which none of us granted it)

codebytere:
Closing as dependabot seems to be picking these up now 🎉

zeke:
[...]

PR #1332, opened 2018-09-19
[wip] fix: ignore package lock

zeke:
The `update-electron-dependencies` script is not installing devDependencies. Removing the package lock file ensures that the script is doing a fresh npm install from scratch.

🍐 @vanessayuenn


zeke:
This turned out to be a red herring in our investigation of #1323

PR #1338, opened 2018-09-19
chore: remove update-deps script in favor of Dependabot

zeke:
@vanessayuenn are testing out [Dependabot](https://dependabot.com) as an alternative to running our own dependency update script. Dependabot has a new feature that allows modules to be whitelisted for immediate merge once the bot's PR has passed all its status checks.

Once we've verified it's working, we can remove this script.


zeke:
That. Is. Glorious.

PR #1376, opened 2018-10-01
feat: show nightly release on main page

MarshallOfSound:
Depends on https://github.com/electron/releases/pull/26

* Adds the nightly release to the main page
* Makes the releases expand on hover so that the full nightly version number is visible
* Makes the `npm install` instructions easily copy-pastable by making surrounding text non-selectable


MarshallOfSound:
@ckerr Yes, I updated the version of a dependency that was still in the requested range.  The lock was updated but not then package.json itself.

codebytere:
What are your thoughts on maybe changing the threshold for stacking? Such that it stacks as soon as there's clipping of the displayed text? I won't block on it, more curious if you think that'd be better too

MarshallOfSound:
> Such that it stacks as soon as there's clipping of the displayed text?

Determining that with CSS is afaik quite difficult, it varies depending on user font size, display scaling and font width.  To compensate for the fact that some text may be offscreen when you hover over versions they get twice as wide now and have a more focused appearance.

Not sure what the best UX is here but my theory is that this is Good Enough and if someone hates it and has a better idea they'll let us know 😄

PR #1388, opened 2018-10-02
fix: properly transform browserified code from node_modules

BinaryMuse:
I noticed our JS bundle contained a lot of code that used newer, non-transpiled JavaScript features (like arrow functions). The root cause turned out to be the fact that Babel was ignoring code from `node_modules`. However, some dependencies contain code that does not transform properly, so the best solution I can think of is to maintain a list of problematic modules and add them to a blacklist. Luckily, only `lodash` and `prettydate` are problematic (for now).

As part of this fix, I also updated us to the latest version of Babel, including a very large upgrade to `@babel/preset-env`.


middleware/browserify.js
```diff
@@ -1,20 +1,25 @@
...
-const brfs = require('brfs')
 
-function babelifyMiddleware (entries, brOptions, baOptions) {
-  brOptions = brOptions || {}
-  brOptions.transform = brOptions.transform || []
-  baOptions = baOptions || {}
-  baOptions.presets = baOptions.presets || ['babel-preset-env']
-  brOptions.transform.unshift(babelify.configure(baOptions))
-  brOptions.transform.unshift(brfs)
+const nodeModulesToAvoidBabelifying = [
```
zeke: Love this constant name. 👍

PR #1389, opened 2018-10-02
feat: more advanced app search

### LINKED ISSUE #692: Filter out apps by unwanted keywords
I wasn't looking for anything in particular when I checked out [`electron.atom.io/apps`](https://electron.atom.io/apps/); I just wanted to see what sort of stuff people had done.

I noticed how many of these were Markdown editors, so I instinctively punched `NOT:markdown` into the filter-field (I'm probably too used to how GitHub [handles this](https://github.com/search?q=extension%3Ascopes+NOT+nothack&type=Code)...). I tried `-markdown`, but that didn't work either.

So, erm, randomly suggested improvement coming through. =)

BinaryMuse:
This introduces more advanced search functionality based on [lunr](https://lunrjs.com/) (and thus removes the need for List.js). It enables the following functionality:

* **Exclusions**: prefix a search term with `-` to exclude that term.

    electron_apps___electron

* **Boolean AND**: prefix terms with `+` to force them all to be included.

    electron_apps___electron

* **Wildcards**: use `*` to perform a wildcard search

    electron_apps___electron

* **Field search**: use `field:` to search in a specific field

    electron_apps___electron

* **Term boosting**: use `^` to boost a score for a certain term

    electron_apps___electron

    electron_apps___electron

* **Fuzzy searching**: use `~` to enable fuzzy searching

    electron_apps___electron

While not all of these features are likely to be used by end users, they all come for free and we always have the option to use them with a nice JavaScript API instead of obtuse symbols in a search string. In addition, I've changed the way results are shown and hidden when a search is performed, ideally making it faster <sup>~citation~ benchmark needed</sup>.

- [x] Benchmark new list update code
- [x] Compare bundle size before/after change

Depends on #1388
Closes #692


BinaryMuse:
Some quick metrics:

|         | Bundle Size | Search for "markdown" | Reset search box |
|---------|-------------|-----------------------|------------------|
| List.js | 662kb       | 39ms                  | 161ms            |
| lunr    | 674kb       | 31ms                  | 29ms             |
| Delta | +12 kb (<2% larger) | -8ms (20% faster) | -132ms (82% faster) |

zeke:
I'm happy with shipping this PR, but want to mention that we also have all this app data indexed in Algolia. So we could eventually drop both lunr and List.js in favor of Algolia's autocomplete.js which we're already depending on on every page (because the search is in the header).

scripts/create-filter-list.js
```diff
@@ -1,8 +1,106 @@
...
+    this.search(this.input.value)
+  }
+
+  search (text) {
+    try {
+      const results = this.index.search(text)
+      this.updateList(results)
+    } catch (err) {
+      if (err.name === "QueryParseError") {
+        // usually because the user is typing special symbols; safe to ignore
```
ckerr: :+1:

PR #1449, opened 2018-10-22
feat: make the core site work offline

MarshallOfSound:
Service workers are fun 👍


views/layouts/main.html
```diff
@@ -13,6 +13,11 @@
     <div id="refinement-list"></div>
     {{> footer}}
     {{> anchor_links}}
+    <script>
+    if ('serviceWorker' in navigator) {
+      navigator.serviceWorker.register('/sw.js');
```
zeke: how about naming this file something like `service-worker.js` to make its purpose more obvious?

zeke:
This is cool!

scripts/service-worker.js
```diff
@@ -0,0 +1,46 @@
+/* global importScripts, workbox */
+
+importScripts('https://storage.googleapis.com/workbox-cdn/releases/3.6.1/workbox-sw.js')
+importScripts('/scripts/precache.js')
+
+if (workbox) {
```
zeke: Should this be `typeof workbox !== `undefined``?

zeke:
Can you add a short blurb to the README about this feature, how it works, why it exists, etc?

scripts/service-worker.js
MarshallOfSound: Both work, will update for clarity

MarshallOfSound:
@zeke note added 👍

PR #1465, opened 2018-10-24
fix: reset SW cache on language change

MarshallOfSound:
My assumption that all content was static based on URL was incorrect, turns out our translations are done based on a cookie.  This PR updates our base JS file (which will be invalidated by the SW on reload) with some logic to wipe the SW cache whenever a locale link is followed.

There is probably a better solution with caching based on that cookie but for now this makes locale switching work again

### DIFF
--- a/scripts/reset-lang-cache.js
+++ b/scripts/reset-lang-cache.js
@@ -1,0 +1,18 @@
+module.exports = () => {
+  const links = document.querySelectorAll('[href^="/languages/"]')
+  for (const link of links) {
+    link.addEventListener('click', (e) => {
+      if (window.caches) {
+        e.preventDefault()
+
+        window.caches.keys()
+          .then((keys) => {
+            for (const key of keys) {
+              window.caches.delete(key)
+            }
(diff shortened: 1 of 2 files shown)


scripts/reset-lang-cache.js
```diff
@@ -0,0 +1,16 @@
+module.exports = () => {
+  const links = document.querySelectorAll('[href^="/languages/"]')
+  for (const link of links) {
+    link.addEventListener('click', (e) => {
+      e.preventDefault()
+
+      window.caches.keys()
```
BinaryMuse: Do we need to do any feature detection around this block?

MarshallOfSound: Updated

BinaryMuse:
This looks reasonable to me. So it'll just clear the cache whenever someone changes a language? And then on the next page load it'll refill based on the precache?

MarshallOfSound:
> So it'll just clear the cache whenever someone changes a language? And then on the next page load it'll refill based on the precache?

Yup and yup 👍

PR #1495, opened 2018-11-04
fix: memory leak on feed construction

codebytere:
The feed was previously abstracted out but in such a way that for every request to the release endpoint, all of the releases were added to the feed again. This lead to memory leaks and errors, as seen here:

screen shot 2018-11-03 at 7 14 46 pm

This abstracts out feed population to mitigate those errors.

/cc @MarshallOfSound

### DIFF
--- a/routes/feed/mainFeed.js
+++ b/routes/feed/mainFeed.js
@@ -1,12 +1,13 @@
 const Feed = require('feed').Feed
 const description = require('description')
+const memoize = require('fast-memoize')
 
 const types = {
   blog: 'blog',
   releases: 'releases'
 }
 
-module.exports.setupFeed = (type, items) => {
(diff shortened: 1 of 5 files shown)


routes/feed/mainFeed.js
```diff
@@ -1,16 +1,51 @@
...
+    generator: 'Electron website',
+    feedLinks: {
+      json: 'https://electronjs.org/releases.json',
+      atom: 'https://electronjs.org/releases.xml',
+      json1: 'https://electronjs.org/blog.json',
+      atom1: 'https://electronjs.org/blog.xml'
+    }
+  })
+
+  if (type === types.releases) {
```
MarshallOfSound: maybe switch here for clarity? `switch (type) {`

PR #1519, opened 2018-11-08
fix: support new electron-releases data format

BinaryMuse:
Fixes https://github.com/electron/electronjs.org/issues/1511

Ref: https://github.com/electron/releases/pull/27


zeke:
👏

PR #1606, opened 2018-12-10
chore: added lint as precommit

maddhruv:


codebytere:
hey @maddhruv we temporarily ditched the service worker; would you mind fixing the conflict so we can get this merged? thanks 😁

Issue #1654, opened 2018-12-19
Service worker unable to unregister


HashimotoYT:
The cached webpage is trying to get `./service-worker`, which does not exist, and we cannot unregister the existing `./service-worker`. Need to remove cached files before unregistering service worker.

![default](https://user-images.githubusercontent.com/24681191/50249150-e374fd80-03e5-11e9-922f-eb09e8e2f8a6.png)

/cc @BinaryMuse

BinaryMuse:
Thanks for the catch, @HashimotoYT. I'm not super familiar with this — can you elaborate on "need to remove cached files before unregistering service worker"?

HashimotoYT:
The service worker caches the files for the next offline download, currently the service worker trying to load the cached page and refresh the cache for the next download. We need to delete the cached storage for unregister service worker. I don’t know how to delete cached storage because I don’t work with it.

BinaryMuse:
> I don’t know how to delete cached storage because I don’t work with it.

Hah, yeah, same. I'll try #1655 and go from there. Thank you!

PR #1655, opened 2018-12-19
fix: properly remove caches before removing service worker

BinaryMuse:
Working on fixing #1654. Not super confident about this so 💭 and 👀 appreciated.

/cc @HashimotoYT

### DIFF
--- a/scripts/service-worker.js
+++ b/scripts/service-worker.js
@@ -1,0 +1,15 @@
+self.addEventListener('activate', function (event) {
+  event.waitUntil(
+    caches.keys().then(function (cacheNames) {
+      return Promise.all(
+        cacheNames.filter(function (cacheName) {
+          // Return true if you want to remove this cache,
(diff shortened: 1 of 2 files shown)


HashimotoYT:
I start loading local website with cached files. After I start this implementation for check working.

HashimotoYT:
Not sure, but it looks like it did not help, the script for unregistration did not appear.
default

HashimotoYT:
This only reproducible for cached page, empty load work pretty.

BinaryMuse:
🤔 That's interesting, I would expect that removing `index.html` from the cache would cause the SW to be removed on the next page load

BinaryMuse:
I was able to reproduce locally, and it seems this `skipWaiting` call during `install` takes care of it.

HashimotoYT:
It works, reloading removing all cached files.

PR #1667, opened 2018-12-28
main page, releases block

Arhell:
Hello, on 768
http://joxi.ru/n2YLp5ZToNebvm


BinaryMuse:
@Arhell Would you mind clarifying the reason for this change? Closing this in the meantime to keep things tidy. Thanks!

Arhell:
Yes, at 768 (ipad), in the right corner there is a bar that stretches the screen and the content does not fit the full width, an example is shown in the screenshot above, you can check in the browser for a test

Arhell:
There are other solutions, for example, not to touch the container as a whole, but to work directly with this section.

PR #1669, opened 2018-12-30
fix: strip leading $ from copied blocks

codebytere:
In codeblocks like the following: 
```
$ cd src/electron
$ git remote remove origin
$ git remote add origin https://github.com/electron/electron
$ git branch --set-upstream-to=origin/master
$ cd -
```

the `$` was being copied alongside the commands themselves, which is aesthetically nice but rather unhelpful for being able to immediately execute code contained therein. This PR thus alters the text before it is copied to the clipboard, such that the leading `$` at the beginning of lines are not included in that text. 

h/t @ryzokuken for bringing this to attention


scripts/copy-code-to-clipboard.js
```diff
@@ -1,10 +1,20 @@
...
-const clipboard = new ClipboardJS('.btn-clipboard')
-
-clipboard.on('success', (e) => {
-  e.clearSelection()
+const clipboard = new ClipboardJS('.btn-clipboard', {
+  // strip leading console $ from copied codeblocks
+  text: trigger => {
+    const codeBlockID = trigger.getAttribute('data-clipboard-target').substr(1)
+    let codeBlock = document.getElementById(codeBlockID).textContent
+    console.log(codeBlock)
```
BinaryMuse: 🔥

scripts/copy-code-to-clipboard.js
```diff
@@ -1,10 +1,20 @@
...
+    codeBlock = codeBlock.replace(/\n\$/g, '\n') // eslint-disable-line no-useless-escape
+    if (codeBlock.startsWith('$ ')) {
+      codeBlock = codeBlock.substr(2)
+    }
```
BinaryMuse: How do you feel about changing this whole section to something like

```javascript
codeBlock = codeBlock.replace(/^\$\W*/gm, '')
```

This handles any number of spaces after the `$` and doesn't require the `if` statement to take care of the first line.

PR #1731, opened 2019-01-08
chore: update electron-releases dep

codebytere:
Pick up changes from https://github.com/electron/releases/pull/35.

cc @BinaryMuse


BinaryMuse:
@codebytere Dependabot automatically opened and merged a PR bumping this dep

PR #1832, opened 2019-01-10
fix: truncating for download block

HashimotoYT:
Fixing https://github.com/electron/electronjs.org/pull/1778#issuecomment-452908909

| Before | After |
| --- | --- |
| before | after

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1088,7 +1088,7 @@
         },
         "events": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/events/-/events-1.1.1.tgz",
+          "resolved": "http://registry.npmjs.org/events/-/events-1.1.1.tgz",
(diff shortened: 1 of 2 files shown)


BinaryMuse:
Oh I just noticed the "see all 11 files" at the bottom of the sidebar. Was the original intent for the buttons to clip? 🤔

BinaryMuse:
@HashimotoYT I'm sorry, I didn't mean your original PR, but the original styling for the website. The link to "View all 11 files" only makes sense if we're not already showing all 11 files.

HashimotoYT:
It's true, we don't show all 11 files, we only show platform specific files, how seed below. All files described here displayed by deleting "display: none" style.

name

BinaryMuse:
Ah, I see. Thanks for clarifying.

I think this looks good to me!

PR #1950, opened 2019-01-21
fix: correct import tag_name

### LINKED ISSUE #1941: History page not have version in url
Should have version, but not:

where version

HashimotoYT:
Fixes #1941


views/docs/history.html
```diff
@@ -27,13 +27,13 @@
...
       <td>
-          <a href="https://github.com/electron/electron/blob/{{this.tag_name}}{{../doc.href}}.md">
+          <a href="https://github.com/electron/electron/blob/{{this.data.tag_name}}{{../doc.href}}.md">
             <span class="octicon octicon-mark-github mr-1"></span>
             {{../doc.slug}}.md @ {{this.data.version}}
           </a>
       </td>
       <td>
-          <time data-date="{{this.data.created_at}}" data-format="%B %d, %Y">{{this.created_at}}</time>
+          <time data-date="{{this.data.created_at}}" data-format="%B %d, %Y"></time>
```
BinaryMuse: Thoughts about changing this to `published_at`? I noticed a few cases where they don't match and the `published_at` seems to be the correct date.

```suggestion
          <time data-date="{{this.data.published_at}}" data-format="%B %d, %Y"></time>
```

views/docs/history.html
```diff
@@ -27,13 +27,13 @@
   {{#each doc.releases}}
     <tr>
       <td>
-          <a href="https://github.com/electron/electron/blob/{{this.tag_name}}{{../doc.href}}.md">
+          <a href="https://github.com/electron/electron/blob/{{this.data.tag_name}}{{../doc.href}}.md">
```
BinaryMuse: Oops, my fault 🙈

HashimotoYT:
I changed `created_at` to `published_at`, because it's really show mostly true date. I did not find in the documentation API than these values differ.

BinaryMuse:
> I did not find in the documentation API than these values differ.

Yeah, I don't think they're supposed to. I'll add it to the list :)

PR #2027, opened 2019-01-31
feat: add v8::Handle / Electron 5.0 blog post

BinaryMuse:
Roughed out the skeleton of this post; if anyone has more details or an idea for better instructions to give to our users, I'm all 👂s.

/cc @MarshallOfSound @nornagon @ckerr in particular


data/blog/nodejs-native-addons-and-electron-5.md
```diff
@@ -0,0 +1,23 @@
...
+
+If you're having trouble using a native Node.js addon with Electron 5.0, there's a chance it needs to be updated to work with the most recent version of V8.
+
+---
+
+## Goodbye `v8::Handle`, Hello `v8::Local`
+
+In 2014, the V8 team deprecated `v8::Handle` in favor of `v8::Local` for local handles. Electron 5.0 includes a version of V8 that has finally removed `v8::Handle` for good, and native Node.js addons that still use it will need to be updated before they can be used with Electron 5.0.
+
+The required code change is minimal, but *every* native Node module that still uses `v8::Handle` will fail to build with Electron 5.0 and will need to be modified. The good news is that Node.js v12 will also include this V8 change, so any modules that use `v8::Handle` will need to be updated *anyway* to work with the upcoming version of Node.
```
nornagon: We should probably confirm this with someone on the Node team.

data/blog/nodejs-native-addons-and-electron-5.md
```diff
@@ -0,0 +1,23 @@
...
+
+## I maintain a native addon, how can I help?
+
+If you maintain a native addon for Node.js, ensure you replace all occurrences of `v8::Handle` with `v8::Local`. The former was just an alias of the latter, so no other changes need to be made to address this specific issue.
```
nornagon: This might be a good place to plug `nan` / `napi`, which provide more stable APIs against which to build native modules. cc @MarshallOfSound

data/blog/nodejs-native-addons-and-electron-5.md
BinaryMuse: @nornagon I opted to reword this a little; what do you think of it now?

> The required code change is minimal, but *every* native Node module that still uses `v8::Handle` will fail to build with Electron 5.0 and will need to be modified. The good news is that Node.js will also eventually include this V8 change, so any modules that use `v8::Handle` will need to be updated eventually anyway.

data/blog/nodejs-native-addons-and-electron-5.md
BinaryMuse: I added a small section on N-API

PR #2039, opened 2019-01-31
fix: load static assets from a subdirectory

BinaryMuse:
Currently, we use `express.static` on the root directory. This exposes everything in the repository to the world, which is a security concern. This PR scopes public assets to the `public/` subdirectory, and adds a new URL prefix `/app-img` that can be used to reach into `node_modules/electron-apps/apps/` to get app images.


BinaryMuse:
I poked around the deployment and everything seems to be working as intended. Happy to have folks double check! 😅

BinaryMuse:
I don't feel strongly either way, but my preference is to go ahead and get this merged and rebase them as necessary. I'll take care of #2038 today once this merges.

PR #2087, opened 2019-02-08
fix: 302 redirect www to bare domain

### LINKED ISSUE #2086: 301 redirect to bare hostname
Currently the `www.` subdomain operates as a separate site; we should 301 from the `www.` subdomain to the bare domain.

BinaryMuse:
Redirects requests to the `www.` subdomain to the bare domain via a 301 response.

Fixes #2086


server.js
```diff
@@ -41,6 +42,7 @@ app.engine('html', hbs.express4({
 // Middleware
 app.set('view engine', 'html')
 app.set('views', path.join(__dirname, '/views'))
+app.use(nakedRedirect(true, 'www', 301))
```
MarshallOfSound: Can I request we make this `302` instead of `301` for a few days at least to ensure we aren't redirecting badly.  I'm always scared of `301` because there is no way to undo any damage done by it 😆 **Permanent**

BinaryMuse:
I'm in favor of using the bare domain as the canonical URL instead of the `www.` variant, but I could be persuaded otherwise.

Kilian:
Why specifically the 302 redirect? SEO-considerations?

With regards to https: can/do we want to set [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)? electron

BinaryMuse:
@Kilian 

> Why specifically the 302 redirect?

I had originally used 301 but @MarshallOfSound brought up that it might be a good idea to make sure everything is good to go using 302 first just in case.

> can/do we want to set [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)?

This isn't a bad idea. I want to double-check on how our SSL stuff is set up, but if it looks good I'll open another PR for this.

MarshallOfSound:
Moving the discussion in Slack to this thread.

Turns out we already have HSTS enabled 👍

PR #2088, opened 2019-02-08
feat: add Keyboard Cheatsheet

HashimotoYT:
⚠️ _This continue #1309, which was long was inactive._ ⚠️ 

How it looks:

default

Last things:

- [x] Change description for ⬇️ and ⬆️ 

If it is possible:

~- [ ] Create Cypress e2e test~

Closes https://github.com/electron/electronjs.org/pull/1309


HashimotoYT:
@electron/wg-website it's ready for review.

data/locale.yml
```diff
@@ -14,6 +14,14 @@ nav:
...
   userland: Userland
 
+keyboard_modal:
+  focus_search: Focus search bar.
+  focus_and_clear: Focus search bar and clearns it.
+  scroll_down: Scroll the searched result bottom.
+  scroll_up: Scroll the searched result upper.
+  open: Opens the searched result.
+  open_new_tab: Opens the searched result in new tab.
+
```
BinaryMuse: Some ideas for these phrases:

* Focus search bar
* Focus and clear search bar
* Select the next search result
* Select the previous search result
* Open the selected search result
* Open the selected search result in a new tab

scripts/modal.js
```diff
@@ -0,0 +1,34 @@
...
+  window.addEventListener('keydown', event => {
+    if (event.key === '?') {
+      if (event.target === null || event.target.tagName !== 'INPUT') {
+        modal.style.display = 'block'
+        event.stopImmediatePropagation()
+        event.preventDefault()
+      }
+    }
+
+    if (event.keyCode === 27) {
```
BinaryMuse: How would you feel about including `?` in this, so it could be toggled?

```suggestion
    if (event.keyCode === 27 || event.key === '?') {
```

scripts/modal.js
```diff
@@ -0,0 +1,34 @@
+module.exports = function installModal () {
+  // Get the modal
+  const modal = document.getElementById('myModal')
```
BinaryMuse: How would you feel about using a different ID here? Something like `keyboard-reference-modal` or something?

scripts/modal.js
```diff
@@ -0,0 +1,34 @@
...
+
+  // Get the <span> element that closes the modal
+  const span = document.getElementsByClassName('close')[0]
```
BinaryMuse: I think it would be good to scope this selector:

```suggestion
  const span = modal.getElementsByClassName('close')[0]
```

scripts/modal.js
```diff
@@ -0,0 +1,34 @@
...
+      }
+    }
+
+    if (event.keyCode === 27) {
+      modal.style.display = 'none'
+    }
+  }, true)
+
+  // When the user clicks on <span> (x), close the modal
+  span.onclick = () => {
```
BinaryMuse: Can we use `addEventListener` here or is there a reason to use `onclick` (same question for `window.onclick` below)

HashimotoYT:
1. > Closing the modal with the escape key jumps focus to the search bar

All keys in modal dialog blocked, besides Esc.

2. > Typing `?` in the app search field focuses the search box

Looks like error here: https://github.com/electron/electronjs.org/blob/b03139a679268ac6a94b71dadaa0f23fd9d96ff0/scripts/modal.js#L11
but it expected behavior, very intresing.

3. > Can we use `addEventListener` here or is there a reason to use `onclick` (same question for `window.onclick` below)

Yup, done 💯.

4. > How would you feel about using a different ID here? Something like `keyboard-reference-modal` or something?

Changed to `kb-shortcut-dialog`

5. > How would you feel about including `?` in this, so it could be toggled?

I agreede make this toggleable, but it has one problem we going to infinity loop opening and closing modal, I think it can wait.

BinaryMuse:
> 1. Typing `?` in the app search field focuses the search box

I think this problem is originating from https://github.com/electron/search-with-your-keyboard

HashimotoYT:
@BinaryMuse Is it ready for merge, or more propositions or bugs?

BinaryMuse:
@HashimotoYT I think this is good to go! ✨

BinaryMuse:
Thanks for getting this across the finish line 🏁

PR #2091, opened 2019-02-08
feat: precompile assets on build

### LINKED ISSUE #2084: Precompile static assets on deployment
Serving CSS and JS from in-memory is convenient, but data from Google Analytics suggests it's adding significant load time to our site. We should consider precompiling these assets in Heroku's release phase and serve them from disk.

BinaryMuse:
Currently we compile and serve JS and CSS from memory when those files are requested, resulting in significant delays for many requests. This PR implements precompilation for production mode.

Fixes #2084


script/precompile-assets
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
HashimotoYT: it's not the cross-platform able solution, what about the switch to Node for cross-platform running?

BinaryMuse: That's a good idea. I'll do that today.

BinaryMuse:
I've updated the script to be in JavaScript; I was also able to save about 15% of the bundle size by modifying our lodash imports.

package.json
```diff
@@ -101,6 +101,7 @@
     "standard": "^12.0.1",
     "supertest": "^3.4.2",
     "supertest-session": "^3.3.0",
+    "uglify-js": "^3.4.9",
```
HashimotoYT: What about change `uglify-js` what looks no longer maintained to `terser` who faster?

BinaryMuse: I would be for this, except that in dev mode, `browserify-middleware` uses `uglify-js` internally by default. If we switch this to `terser`, we should also find a way to change it there for consistency. I'll look into it.

HashimotoYT: Made PR `browserify-middleware` (https://github.com/ForbesLindesay/browserify-middleware/pull/126), who changes `uglify-js` to `terser`, if it gets merge, we can move next.

BinaryMuse: That repo isn't seeing a lot of activity, so I'm not certain it'll get merged. As I've been looking and researching, I think I'd ultimately like to remove browserify/browserify-middleware completely and replace it with something else. How do you feel about merging this PR and tackling that in another?

Issue #2102, opened 2019-02-11
Update Markdown renderer


BinaryMuse:
We currently use [hubown](https://github.com/electron/hubdown) for rendering Markdown on the blog and in other places (like the docs when we convert to HTML in the [i18n repo](https://github.com/electron/i18n)). While pretty good, it fails in a few places (e.g. https://github.com/electron/electron/issues/16034). It would be nice to use a Markdown renderer that results in HTML more similar to what one sees on GitHub.

[cmark-gfm](https://github.com/BinaryMuse/node-cmark-gfm) provides Node.js bindings to [libcmark-gfm](https://github.com/github/cmark-gfm), which is what GitHub uses for Markdown. If we replace the Markdown renderer on electronjs.org, we'll want to do an audit to ensure that things like permalinks remain consistent. Some functionality from hubdown, like syntax hightlighting and autolink-headings, isn't readily available in cmark and we would need to find a workaround.

PR #2160, opened 2019-02-18
feat: use cmark-gfm for rendering blog posts

BinaryMuse:
Per #2102, this PR updates the site to use [cmark-gfm](https://github.com/BinaryMuse/node-cmark-gfm) (via [BinaryMuse/electron-markdown](https://github.com/BinaryMuse/electron-markdown)) instead of [hubdown](https://github.com/electron/hubdown) for rendering blog posts (this is the only markdown rendering done directly by the site). The PR also makes a few other refactoring-style changes to the blog route handlers.

* Replaces hubdown with cmark-gfm
* Removes `blog` middleware and replaces it with actual routes that use a new `blog` lib for finding and parsing blog posts
* Modifies blog feed to use the new `blog` lib
* Ensures blog posts are loaded asynchronously; fixes race condition where blog routes could run with no posts loaded
* Moves blog views from `views/posts` to `views/blog`
* Changes Node version from 8 to 10

Todo:

- [x] Add IDs to headings
- [x] Turn headings into links
- [x] Syntax highlighting
- [x] Fix extra space at bottom of code listings
- [x] Fix iframes (e.g. YouTube videos on `/blog/2015-whats-new-in-electron`)


BinaryMuse:
Okay, I believe this is feature-complete when compared with the previous renderer. I still want to do some comparisons of the old and new results to make sure nothing unexpected broke.

BinaryMuse:
@electron/wg-website This is ready for review.

I downloaded a copy of every blog post on the current site and on this PR's deployment, and generated a diff for each. [The results show](https://gist.github.com/BinaryMuse/28a41a914149d34a61564fa46b0a2d37) that the changes are either (1) whitespace changes (particularly an extra non-visible newline at the end of code blocks), (2) single/double quote changes, and (3) some `<li>` elements no longer have their contents wrapped in `<p>`.

PR #2187, opened 2019-02-22
fix: Using search parameter 'q' instead of 'query'

### LINKED ISSUE #1284: Use `q` param instead of `query` for searches
Shorter. Like Google.

https://community.algolia.com/instantsearch.js/v2/guides/routing.html may be the ticket.

Arp-1:
This PR resolves #1284


ckerr:
@Arp-1 am I reading this PR correctliy? This looks like a dupe of #2186, which has already been merged. I think this PR can be closed.

Arp-1:
@ckerr actually I opened that [PR](https://github.com/electron/electronjs.org/pull/2186) but that was having CI tests failed so I closed that and opened this one.

zeke:
Thanks for the PR @Arp-1!

Hey @haroenv 👋 -- Is this the Algolia-recommended way to accomplish this?

PR #2219, opened 2019-02-28
feat: make manual copy-paste of bash commands nice

MarshallOfSound:
This makes prefixed "$" signs in bash blocks unelectable.

It also makes bash comments and their trailing newline char unelectable

### DIFF
--- a/scripts/copy-code-to-clipboard.js
+++ b/scripts/copy-code-to-clipboard.js
@@ -21,6 +21,19 @@ module.exports = function copyCodeToClipBoard () {
     button.setAttribute('data-clipboard-target', `#${id}`)
     button.innerHTML = window.localized.clipboard.copy
     code.parentElement.appendChild(button)
+
+    // Make prefixed dollars unselectable
+    code.innerHTML = code.innerHTML.replace(/\n\$ /g, '\n<span class="no-select">$ </span>')
+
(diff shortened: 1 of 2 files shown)


scripts/copy-code-to-clipboard.js
```diff
@@ -21,6 +21,19 @@ module.exports = function copyCodeToClipBoard () {
     button.setAttribute('data-clipboard-target', `#${id}`)
     button.innerHTML = window.localized.clipboard.copy
     code.parentElement.appendChild(button)
+
+    // Make prefixed dollars unselectable
+    code.innerHTML = code.innerHTML.replace(/\n\$ /g, '\n<span class="no-select">$ </span>')
```
ckerr: If the block begins with a `$`, this regex won't match on it.

Maybe `(?:^\n)` or `\b` instead of `\n`?

MarshallOfSound: Could catch, I used `(^|\n)` below so I'll use that here as well

ckerr: Even better. Now that I think about it, `\b` would have unintended side-effects and I left an `|` out of my first suggestion.

One suggestion, making it a non-capture group is marginally more efficient and we're not using the capture -- so `(?:^|\n)` seems like it would be the best of both worlds

PR #2242, opened 2019-03-06
wip feat: add primer version checker

HashimotoYT:
This PR adds script what checks if we behind of Primer what published to npm.

For more information read this note: https://hackmd.io/zZiYBmJuRsqTyDdrwluqbQ

/cc @electron/wg-website


BinaryMuse:
@HashimotoYT Can you clarify on this point?

> ## Why not bump it by add @primer/css?
> 
> Looks now Primer uses a single package instead of small packages, we can simply add `@primer/css` and take a needed style. But it still looks very risky, mostly if we have an error when updating `@primer/css`, we unable to fix it by updating styles locally. For now, I think we need manually port styles to repository instead of using npm version (but I'm open for discussion 💬).

Do we have a history of Primer breaking in such a way that would require us to completely replace one of its `.scss` files? If not, I'm strongly in favor of just installing `@primer/css` via npm and using it directly. The more of these check-and-update scripts we have (see also the `generate-octicons` script), the more complex the project becomes and the likely it is that we'll forget to check and update something.

(Also, even if we use `@primer/css` from npm, we can always copy over styles into the `public` folder if we really need to.)

PR #2277, opened 2019-03-11
feat: create Governance landing page

HashimotoYT:
Based on top level readme.

Fixes https://github.com/electron/governance/projects/1#card-18085872


codebytere:
@HashimotoYT I think we'll want to surface this on the front page somewhere clear? In the top bar next to Releases, perhaps?

HashimotoYT:
1. I agree to remove information about security and move this to `/security` page in some other time. I'm not agreed to create an individual page for each group because changes should from `e/governance` be mirrored manually to this eaches pages.

This answer resolve this question @MarshallOfSound?

MarshallOfSound:
> I'm not agreed to create an individual page for each group because changes should from e/governance be mirrored manually to this eaches pages.

I agree, we should find a way to sync that content rather than mirror it.  We have infra for `e/e`'s docs folder we can probably do the same here.

That said it resolves my point for now 👍 

I just pushed up a commit that adds some fancy octicons and makes the whole page responsive so folks can view it easily on mobile 🎉

BinaryMuse:
https://github.com/electron/electronjs.org/pull/2304 has been merged into this branch.

PR #2307, opened 2019-03-17
fix: updated instantsearch to v3.1.1

Arp-1:
@HashimotoYT , I have updated instantsearch to v3.1.1 and I only needed to change `/scripts/search.js` file(and not any styles) but the margin between github icon and search bar is decreased(which I don't understand why :confused: ). Could you please guide me to fix it :smiley: .
I've attached the appropriate images.

Image before applying the changes

Image after applying the changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   "dependencies": {
     "@babel/core": "^7.3.4",
     "@babel/preset-env": "^7.3.4",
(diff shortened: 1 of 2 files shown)


molant:
We've migrated the search to a different system and this is no longer needed.

Thanks for the work @Arp-1!

PR #2314, opened 2019-03-19
docs(blog): add from-native-to-js

codebytere:
Part of a hopefully ongoing series about how some of Electron's internal mechanisms work in more detail.

This post explains how Electron's features written in C++ or Objective-C get to JavaScript so they're available to an end-user.

cc @ckerr


data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+.SetProperty("property_name", &getter_function_to_bind)
+```
+
+or
+
+```cpp
+.SetProperty("property_name", &getter_function_to_bind, &setter_function_to_bind)
+```
+
+These would in turn be the JavaScript implementations of [Object.defineProperty](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty):
```
MarshallOfSound: Not sure if the technicality of this needs to be this specific but `SetProperty`'s equivalent in JS land would be this

```js
function App {}
Object.definiteProperty(App.prototype, 'myProperty', {
  get() {
    return _myProperty
  }
})
```

Notably defining on the prototype instead of a property on an instance

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+date: '2018-03-19'
+---
+
+How do Electron's features written in C++ or Objective-C get to JavaScript so they're available to an end-user?
+
+---
+
+## Background
+
+[Electron](https://electronjs.org) is a JavaScript platform whose primary purpose is to lower the barrier to entry for developers to build robust desktop apps. However, at its core, Electron doesn't remove the need for platform-specific functionality to be written in a given system language.
```
MarshallOfSound:
```suggestion
[Electron](https://electronjs.org) is a JavaScript platform whose primary purpose is to lower the barrier to entry for developers to build robust desktop apps without worrying about platform-specific implementations. However, at its core, Electron itself still needs platform-specific functionality to be written in a given system language.
```

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+
+In reality, Electron handles the native code for you so that you can focus on a single JavaScript API.
+
+How does that work, though? How do Electron's features written in C++ or Objective-C get to JavaScript so they're available to an end-user?
+
+To trace this pathway, start with the [`app` module](https://electronjs.org/docs/api/app).
```
MarshallOfSound: Not sure if this is the intended tone of this post but it felt like it fitted in with the "walk through" style of the post

```suggestion
To trace this pathway, let's start with the [`app` module](https://electronjs.org/docs/api/app).
```

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+
+By opening the [`app.ts`](https://github.com/electron/electron/tree/master/lib/browser/api/app.ts) file inside our `lib/` directory, you'll find the following line of code towards the top:
```
MarshallOfSound: Maybe link to an exact commit hash here so that the link makes sense for future visitors

MarshallOfSound: Also probably worth noting somewhere in this post that this is The Way Things Works as of now and could change in the future if you're reading this blog post a year on 👍

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+
+```js
+const binding = process.electronBinding('app')
+```
+
+This line points directly to Electron's mechanism for binding its C++/Objective-C modules to JavaScript for use by developers. This function is created by the header and [implementation file](https://github.com/electron/electron/tree/master/atom/common/api/atom_bindings.cc) for the `ElectronBindings` class.
```
MarshallOfSound: This is now `electron_bindings.cc` right?

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+In the above line, `.SetMethod` is called on `mate::ObjectTemplateBuilder`. `.SetMethod` can be called on any instance of the `ObjectTemplateBuilder` class to set methods on the [Object prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/prototype) in JavaScript, with the following syntax:
+
+```cpp
+.SetMethod("method_name", &function_to_bind)
+```
+
+This is the JavaScript equivalent of:
+
+```js
+app.prototype.getGPUInfo = function () {
```
MarshallOfSound:
```suggestion
function App{}
App.prototype.getGPUInfo = function () {
```

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+  get() {
+    return _myProperty
+  }
+})
+```
+
+and
+
+```js
+Object.defineProperty(app, 'myProperty', {
```
MarshallOfSound: Same here RE defining on the prototype

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+  }
+  set(newPropertyValue) {
+    _myProperty = newPropertyValue
+  }
+})
+```
+
+With this, it’s possible to create JavaScript objects formed with prototypes and properties as developers expect them, and more clearly reason about functions and properties implemented at this lower system level!
+
+The decision around where to implement any given module method is itself a complex and oft-nondeterministic one, which I'll save for a later post.
```
MarshallOfSound: Just so you don't pre-dob yourself in for writing it 😆 Feel free to dismiss this if you do want the original wording 👍 

```suggestion
The decision around where to implement any given module method is itself a complex and oft-nondeterministic one, which we'll cover in a future post.
```

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
+---
+title: From Native to JS in Electron
```
felixrieseberg: I'd suggest making the title a bit sharper – elaborate on `native` and use `JavaScript` instead of JS.

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+
+## `native_mate`
+
+Answers to this question can be found in `native_mate`:  a fork of Chromium's [`gin` library](https://chromium.googlesource.com/chromium/src.git/+/lkgr/gin/) that makes it easier to marshal types between C++ and JavaScript.
+
+Inside `native_mate/native_mate` there's a header and implementation file for `object_template_builder`. This is what allow us to form modules in native code whose shape conforms to what JavaScript developers would expect.
+
+### `mate::ObjectTemplateBuilder`
+
+By considering every module in Electron to be an object, the mechanism behind using an object template creator becomes clearer. V8 implements the JavaScript (ECMAScript) specification, so its native functionality implementations can be directly correlated to implementations in JavaScript. For example, [`v8::ObjectTemplate`](https://v8docs.nodesource.com/node-0.8/db/d5f/classv8_1_1_object_template.html) gives us JavaScript objects without a dedicated constructor function and prototype. It uses `Object[.prototype]`, and in JavaScript would be equivalent to [`Object.create()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create).
```
felixrieseberg:
```suggestion
By considering every module in Electron to be an `object`, the mechanism behind using an object template creator becomes clearer. V8 implements the JavaScript (ECMAScript) specification, so its native functionality implementations can be directly correlated to implementations in JavaScript. For example, [`v8::ObjectTemplate`](https://v8docs.nodesource.com/node-0.8/db/d5f/classv8_1_1_object_template.html) gives us JavaScript objects without a dedicated constructor function and prototype. It uses `Object[.prototype]`, and in JavaScript would be equivalent to [`Object.create()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create).
```

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+
+To see this in action, look to the implementation file for the app module: [`atom_api_app.cc`](https://github.com/electron/electron/tree/master/atom/browser/api/atom_api_app.cc) At the bottom is the following:
```
felixrieseberg:
```suggestion
To see this in action, look to the implementation file for the app module, [`atom_api_app.cc`](https://github.com/electron/electron/tree/master/atom/browser/api/atom_api_app.cc). At the bottom is the following:
```

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,103 @@
...
+  get() {
+    return _myProperty
+  }
+  set(newPropertyValue) {
+    _myProperty = newPropertyValue
+  }
+})
+```
+
+With this, it’s possible to create JavaScript objects formed with prototypes and properties as developers expect them, and more clearly reason about functions and properties implemented at this lower system level!
```
felixrieseberg: I'd suggest either cutting `With this` or replacing it with what you exactly mean by `this`. Otherwise people might have a typical JavaScript issue – figuring out what scope you're actually talking about 😛

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,106 @@
...
+
+At present, answers to this question can be found in `native_mate`:  a fork of Chromium's [`gin` library](https://chromium.googlesource.com/chromium/src.git/+/lkgr/gin/) that makes it easier to marshal types between C++ and JavaScript.
+
+Inside `native_mate/native_mate` there's a header and implementation file for `object_template_builder`. This is what allow us to form modules in native code whose shape conforms to what JavaScript developers would expect.
+
+### `mate::ObjectTemplateBuilder`
+
+By considering every module in Electron to be an `object`, the mechanism behind using an object template creator becomes clearer. V8 implements the JavaScript (ECMAScript) specification, so its native functionality implementations can be directly correlated to implementations in JavaScript. For example, [`v8::ObjectTemplate`](https://v8docs.nodesource.com/node-0.8/db/d5f/classv8_1_1_object_template.html) gives us JavaScript objects without a dedicated constructor function and prototype. It uses `Object[.prototype]`, and in JavaScript would be equivalent to [`Object.create()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create).
```
felixrieseberg: I don't have a good proposal, but just FYI: I had to read this sentence three times before I fully got it 😅

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,106 @@
...
+
+To see this in action, look to the implementation file for the app module, [`atom_api_app.cc`](https://github.com/electron/electron/tree/master/atom/browser/api/atom_api_app.cc). At the bottom is the following:
```
felixrieseberg: For longevity, I'd recommend linking to a hard commit in case we ever move the file 🔗

data/blog/from-native-to-js.md
```diff
@@ -0,0 +1,106 @@
...
+  }
+  set(newPropertyValue) {
+    _myProperty = newPropertyValue
+  }
+})
+```
+
+It’s possible to create JavaScript objects formed with prototypes and properties as developers expect them, and more clearly reason about functions and properties implemented at this lower system level!
+
+The decision around where to implement any given module method is itself a complex and oft-nondeterministic one, which we'll cover in a future post.
```
felixrieseberg: `oft-nondeterministic`: Is that a typo or a cool word I haven't seen before?

codebytere: it's valid english 😆

data/blog/from-native-to-js.md
codebytere: updated!

PR #2326, opened 2019-03-21
ci: Update Travis to use Node 11

HashimotoYT:
We use Node 10 on prod, why not use 11 on CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "8"
+  - "11"
 notifications:
   email: false
 before_install:


BinaryMuse:
Any reason for 11 instead of 10 to match prod?

HashimotoYT:
Be to one step ahead of Node LTS, it's a good reason to use 11?

BinaryMuse:
🤔 I'm not sure I agree (but I'm willing to be convinced otherwise :)

My thinking is that if we're running on Node 10 on prod, we should use the same version on CI as that is most likely to alert us to relevant issues before we merge and deploy.

PR #2346, opened 2019-03-26
feat: update homepage to `@primer/css`

HashimotoYT:
## How it looks:

| Before | After |
| --- | --- | 
| electronjs org_ | localhost_5000_ |


BinaryMuse:
I think the loss of vertical negative space is even more visible on mobile:

image

HashimotoYT:
@BinaryMuse Maybe it's related only to this `master` backport PR. It has some overrides from Bower styles side. This spaces should be different in global migration PR. Otherwise, I recheck it's when making rebase.

PR #2395, opened 2019-04-04
fix: persist cookie for language choice

codebytere:
Fixes https://github.com/electron/electronjs.org/issues/2394.

Adds back a max cookie age so that the language choice isn't forgotten after browser is closed and reopened.


BinaryMuse:
It looks like this was removed in 5cdae2054f1da6c7cd49eda1b6d0a151a8b07c6d; I'm curious as to the reason this was done — @zeke, do you have any insight to share?

If we reinstate the expiry, do you think we should pick something longer than 24 hours? For a cookie like this that doesn't affect auth or identity, I think we could go with a long-expiring cookie.

zeke:
I forget what the rationale was behind removing the expiry... 🤔 

In any case, I'm okay with setting an expiry and agree with @BinaryMuse that something longer than a day makes sense.

Maybe not a blocker for this PR, but the website should have a cookie notice to be GDPR compliant. I don't know the specific requirements, but I think at the very least the site needs to somehow notify visitors that cookies are being used.

server.js
```diff
@@ -58,6 +58,7 @@ app.use(requestLanguage({
   languages: Object.keys(i18n.locales),
   cookie: {
     name: 'language',
+    options: {maxAge: 24 * 60 * 60 * 1000},
```
BinaryMuse: What do you say to bumping this to a month?

```suggestion
    options: {maxAge: 30 * 24 * 60 * 60 * 1000},
```

codebytere: sounds good to me!

Kilian:
If I may, as someone from the EU that has spent considerable time talking about and implementing measure for GDPR with clients; GDPR in itself doesn't really have anything to say about cookies, just the way you process people's personal data. We use Google Analytics. This is fine, even without notice, if we:

* anonymize the IP
* Don't allow sharing with other Google entities + sign the gdpr agreement with Google. 

This way we don't process personal data (IP is considered personal) so we're good. The first we can add in the javascript, and the other in the analytics admin. I don't see any other tracking, so legally we do not need a cookie policy or notification. It can't *hurt* to add a cookie policy page mentioning basically stating "We use anonymous cookies for site settings", and maybe Github legal would prefer you did.

zeke:
Thanks for the info @Kilian!

PR #2423, opened 2019-04-09
feat: Update releases page to `@primer/css`

HashimotoYT:
| Before | After |
| --- | --- |
| electronjs org_releases_stable | electron-website-pr-2423 herokuapp com_releases_stable |

### DIFF
--- a/public/styles/_releases.scss
+++ b/public/styles/_releases.scss
@@ -1,30 +1,9 @@
 
 // Releases ------------------------------
 
-.vertical-middle {
-  vertical-align: middle !important;
-}
-
-.releases-tag {
-  display: inline-block;
-  margin: 0.3em 0.1em;
-  padding: 0.1em 0.4em;
-  font-size: 0.86em;
-  color: $main-color-subtle;
-  border: 1px solid $border-color;
-  border-radius: $border-radius;
-
-  &:hover {
-    text-decoration: none;
-    color: $main-color-strong;
-    background-color: $main-bg-color-shade !important;
-  }
-}
-
-.r-class {
-  // Magic numbers.
-  margin-right: 128px;
-  margin-left: 128px;
+#releases {
+  margin-right: 7.5%;
+  margin-left: 7.5%;
 }
 
 .release-version-filter .active {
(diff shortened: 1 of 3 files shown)


BinaryMuse:
I feel like this page is quite wide and leaves a lot of empty space in the center of the screen. Thoughts?

image

HashimotoYT:
Container changed because when viewing very long patches, the page looks small scall (strange word, I'm unable to translate this clearer). I'm fine to change this back to `xl`, and relook that on the redesign in other time.

public/styles/_releases.scss
```diff
@@ -13,75 +17,87 @@
...
+    @include breakpoint(md) { font-size: $h2-size !important; }
+  }
+
+  h4 {
+    font-size: $h3-size-mobile !important;
+    font-weight: $font-weight-light !important;
+
+    @include breakpoint(md) { font-size: $h3-size !important; }
+  }
+
```
erickzhao, line 100: Can get rid of all the `!important` if we wrap this entire block in `.markdown-body`.

public/styles/_releases.scss
```diff
@@ -96,7 +112,7 @@
       &:hover {
         text-decoration: none;
         color: $main-color-strong;
-        background-color: $main-bg-color-shade;
+        background-color: $main-bg-color-shade !important;
```
erickzhao, line 115: Unnecessary `!important`

public/styles/_releases.scss
```diff
@@ -117,7 +133,7 @@
 @media (min-width: 700px) {
   .r-resp-header,
   .r-resp-header-toggle {
```
erickzhao, line 135:
```suggestion
  .btn-link.r-resp-header-toggle {
```

Can remove `!important` afterwards

public/styles/_releases.scss
```diff
@@ -117,7 +133,7 @@
 @media (min-width: 700px) {
```
erickzhao, line 133: Can we use Primer breakpoint variables to standardize things?

public/styles/_releases.scss
```diff
@@ -166,6 +182,6 @@
   }
 
   .r-menu {
-    display: none;
+    display: none !important;
```
erickzhao, line 185: Removing this `!important` still works for me

public/styles/_releases.scss
HashimotoYT: I updated that to primer min-width, but looks like them doesn't have breakpoints for `@media`

public/styles/_releases.scss
```diff
@@ -13,75 +17,87 @@
...
+    margin-bottom: 1em !important;
+    margin-left: 1.5em !important;
+    padding-left: 0;
+    padding-top: 0;
+  }
 }
 
 .releases {
-  width: 1012px;
+  width: 1280px;
```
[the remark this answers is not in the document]
HashimotoYT: This now removed and we only set the margin to `7.5%` for left and right.

HashimotoYT:
Well, my final word it's just to close that since I'm really idk what I should do with that now, because it's now starts to be harder.

PR #2577, opened 2019-05-18
refactor: add octicons helper to use his from svg

HashimotoYT:
This PR replaces the old icon font (CSS) way by new via SVG.

## Background of this change

The main issue what makes me start this it's trying to install the fresh website on Node 12 and see how it's unable to compile one native module whats named `ttf2woff2` (native module whats used to build icon font, we do not use this when generating but it needs upstream).

## Why not update this dependency?

It's easy to say, but not if module what required this deprecated and archived. Usage of dependency tree (comments show why and where it's needed):

```ls
electronjs.org@2.0.0 M:\GitHub\electron\electronjs.org
`-- icon-font-generator@2.1.10 (cli wrapper for `webfonts-generator`, also have some cool needed stuff)
  `-- webfonts-generator@0.4.0 (main module what archived, and unable to update module below)
    `-- ttf2woff2@2.0.3 (native module what's not able to build on Node 12 (module alive, and itself updated to Node 12))
```

Another way how we can update this dependency it's made `npm install ttf2woff2 --dev` whats regenerate lock file, and use this VERY BAD WAY (bad guy, duh).

Another reason why this changes it's very large update guide, we unable to update Octicons without user's hands. This contains 6 steps what need to do for an update this. Also, this doesn't work on the Windows platform (this broken when we try to add types flag `--types [svg, ttf]`, this needs to complete steps above, not sure why maybe bad parsing)

## New way

Very more reasons above say this need to change. This PR changes this to use SVG instead of the icon font. :tada:

> Please notice I there compare this realization with React components.

Previously we use span element with class names:

```html
<span class="octicon octicon-alert" />
```

From now we use Handlebars helper what replaces all this:

```html
{{octicon name='alert'}}
```

And the diff how it's changed:

```diff
- <span class='octicon octicon-alert'></span>
+ {{octicon name='alert'}}
```

### How to use this?

Rough says it looks like React class component, for show octicon you need to use the example below, where name is the name of octicon. Please know the name of octicon is a required field.

```html
{{octicon name='alert'}}
```

Also, this has additional parameters such as className, aria-label, width and height.

```html
{{octicon name='alert' className='ahtung'}}
// => <svg ...args class="octicon octicon-alert ahtung" />
```
You can use options singly or alls.

```html
{{octicon name='alert' className='ahtung' width='64' ariaLabel='karamba'}}
// => <svg ...args class="octicon octicon-alert ahtung" width="64" height="64" aria-label="karamba" role="img">
```

Notice that the width populates also to the height or populates match size, also vice versa or you can use these two props together.

And to end how it looks if this is React component (for fun, no more):

```tsx
<Octicon
  name="alert"
  className="ahtung"
  width={64}
  ariaLabel="karamba"
/>
```

### How to replace old usage?

For replacing standard using you can you this example:

```diff
- <span class="octicon octicon-repo-forked pr-3">
+ {{octicon name="repo-forked" className="pr-3"}}
```

Please notice that we no more have `mega-octicon` class. For updating `mega-octicon` class, see the example below of how to replace these:

```diff
- <span class="mega-octicon octicon-repo-forked pr-3">
+ {{octicon name='repo-forked' width='32' height='32' className='pr-3'}}
```

This not great example, because the required width and height you should find manually. Here not other ways to update this, because of GitHub, doesn't more use `.mega-octicon` class.


patched-modules/octicons.js
```diff
@@ -0,0 +1,81 @@
...
+      if (options['width'] || options['height']) {
+        attrObj['width'] = options['width']
+          ? options['width']
+          : (parseInt(options['height']) * data[key].options['width']) / data[key].options['height']
+        attrObj['height'] = options['height']
+          ? options['height']
+          : (parseInt(options['width']) * data[key].options['height']) / data[key].options['width']
+      } else {
+        attrObj['height'] = data[key].width
+        attrObj['width'] = data[key].height
```
ckerr: It looks like `options['width']`, `options['height']`, `options['class']`, etc. are set in the "Set all the default options" section below -- are these 'else' clauses needed?

And if so, is there any alternative to patching? For example wrap our use of `.toSVG()` with something that fills in the defaults at that point in our own files? It will probably cause us less work over time (e.g. when octicons_node has new releases) if we don't use a patched version.

package.json
```diff
@@ -78,6 +77,8 @@
     "primer-support": "^5.0.0",
     "query-string": "^6.5.0",
     "require-dir": "^1.1.0",
+    "octicons": "^8.5.0",
+    "node-sass": "^4.12.0",
```
ckerr: Is the node-sass line needed?

server.js
```diff
@@ -27,6 +28,24 @@ process.env.HOST = process.env.HOST || `http://localhost:${port}`
 
 // Handlebars Templates
 hbs.registerHelper(lobars)
+
+/**
+ * Handlebars helper whats accepts options from the `{{octicon}}` tag,
```
ckerr:
```suggestion
 * Handlebars helper that accepts options from the `{{octicon}}` tag,
```

public/styles/octicons/octicons.scss
```diff
@@ -1,9 +1,11 @@
 @font-face {
   font-family: "octicons";
-  src: url("./octicons.ttf?dfe556e6e1ad2136cdea11cd7db7b84d") format("truetype"),
-url("./octicons.svg?dfe556e6e1ad2136cdea11cd7db7b84d#octicons") format("svg");
+  src: url("./octicons/octicons.ttf?dfe556e6e1ad2136cdea11cd7db7b84d") format("truetype"),
+url("./octicons/octicons.svg?dfe556e6e1ad2136cdea11cd7db7b84d#octicons") format("svg");
 }
 
+@warn "Using octicons from class name is depracted, switch to octicons helper.";
```
ckerr:
```suggestion
@warn "Using octicons from class name is deprecated, switch to octicons helper.";
```

middleware/register-octicons.js
```diff
@@ -0,0 +1,31 @@
+// @ts-check
+
+const octicons = require('../patched-modules/octicons')
+
+/**
+ * Helper function, what accepts name, and other options and return the string
```
ckerr:
```suggestion
 * Helper function that accepts options (e.g. name) and returns a string
```

package.json
HashimotoYT: Needs this to handle version in a lock file for `node-sass-middleware`, without manually handling version of `node-sass` it's hard to updates to newer versions of Node. `node-sass-middleware` defaults version of `node-sass` it's `^4.9.0` (Node 8-10).

patched-modules/octicons.js
HashimotoYT: 5e9f9562d550fd3b1af74d08a520bc11e437e8e6 adds hack to delete `null`'s options for `.toSVG()`.

views/home.html
```diff
@@ -1,4 +1,12 @@
 <main class="homepage PRIMER-REMOVE-ME">
+  <div>
```
BinaryMuse: 🔥

PR #2623, opened 2019-05-31
feat: Improve the docs section

felixrieseberg:
This pull request reworks the documentation section to better guide visitors and users. As discussed and approved in the working group meeting, we'd like to bring some structure to how we list our documentation.

Luckily, I've done something similar for the main entry point into the `docs` folder many moons ago, which has the following benefits:
• Already translated!
• That structure already underwent some discussion and review

It also has a downside: A few links are missing. That's something I'll fix over in `electron/electron`.

localhost_5000_docs (1)


data/locale.yml
```diff
@@ -30,10 +30,15 @@ keyboard_modal:
 docs:
   title: Electron Documentation
   description: See all of the <a href="/docs/all">docs on one page</a> or check out the <a href="/docs/faq">FAQ</a>.
-  guides: Guides
+  guides: "Guides: Get Started with Electron"
   api_reference: API Reference
   api_structures: API Structures
   advanced: Advanced
+  detailed: Detail Guides
```
malept:
```suggestion
  detailed: Detailed Guides
```

data/locale.yml
```diff
@@ -30,10 +30,15 @@ keyboard_modal:
...
+  custom_dom: Custom DOM Elements
+  main_proc_modules: Main Process Modules
+  renderer_proc_modules: Renderer Process Modules
+  both_proc_modules: Modules for both Process Types
```
nornagon:
```suggestion
  both_proc_modules: Modules for Both Process Types
```

data/locale.yml
```diff
@@ -30,10 +30,15 @@ keyboard_modal:
 docs:
   title: Electron Documentation
   description: See all of the <a href="/docs/all">docs on one page</a> or check out the <a href="/docs/faq">FAQ</a>.
-  guides: Guides
+  guides: "Guides: Get Started with Electron"
```
BinaryMuse: 💭 We use a lot of present-tense verbs in our titles and headings; what do you think about "Getting Started" instead?

felixrieseberg:
Alright, I think this is good to go! There's a [matching PR over in electron/electron](https://github.com/electron/electron/pull/18578), but I don't think that we need to wait for it.

data/locale.yml
```diff
@@ -30,10 +30,15 @@ keyboard_modal:
 docs:
   title: Electron Documentation
   description: See all of the <a href="/docs/all">docs on one page</a> or check out the <a href="/docs/faq">FAQ</a>.
-  guides: Guides
+  guides: "Guides: Getting Started with Electron"
   api_reference: API Reference
   api_structures: API Structures
   advanced: Advanced
+  detailed: Detailed Guides
```
Kilian: I know it's already been changed but this title breaks over two lines (as the only one). Can we do "In Detail" instead?

Selection_004

public/styles/_docs.scss
```diff
@@ -189,3 +190,48 @@
...
+
+.docs-guides {
+  h2 {
+    // Match the default from primer
+    margin-bottom: 12px;
+    padding-bottom: 12px;
+  }
+
+  .docs-guides-list {
+    column-count: 3;
```
Kilian: The 4 column list below this one uses classes to only switch to 4 columns at >992px wide, can we similarly add a media query around this one?

```css
@media (min-width: 922px) {
  .docs-guides-list {
    column-count: 3;
  }
}
```

data/locale.yml
malept: If we do that (which I'm fine with), we should get rid of that phrase from the subheaders.

ckerr: At the risk of bikeshedding -- 'Guides' seems more expressive to me than 'In Detail'

malept: Having "Guides: Get Started with Electron" followed by "Guides" without a modifier seems a little odd to me.

ckerr: You're right, it does. Maybe "Detail" would be better here after all.

TBH though I think this is a symptom of a larger problem, that there is a lot of arbitrary organization here. For example, "Debugging the main process" is under "Getting Started", but "Debugging on Windows" is under Advanced".

That's out-of-scope for this PR, but IMO that kind of organization is something we should do in a followup

felixrieseberg: True, I agree strongly! I want to tackle that next, but I'll need to do it over in `electron/electron`.

felixrieseberg: Let's go with `In Detail` as the last change and make additional changes incrementally, otherwise I fear that we'll end up bikeshedding us to death 😅

PR #2646, opened 2019-06-07
fix: add overflow-auto for homepage code block

### LINKED ISSUE #2599: Horizontal overflow bug on homepage
Text inside preformatted block is overflows and visible outside of html page

image

HashimotoYT:
## Why not `white-space: pre-wrap`.

> Fixup from twitter.

Because:

oy see this prewrap

## Why not `white-space: pre-line`.

> Fixup from other pr

Because:

it's preline

## Why this?

Because I'm don't want to fix the bug for mobiles and create a new bug for desktops. I know this issue, I'm the creator of this issue. It's global file what uses not only on the homepage when I'm found the best option to fix this I fix this globally, but this fixes this strange global message.

Life is Strange 2

Closes #2629 
Closes #2599 
Closes #2600

### DIFF
--- a/views/home.html
+++ b/views/home.html
@@ -190,7 +190,7 @@
   </h3>
   <p>{{{localized.quick_start.description}}}</p>
 
-  <figure class="highlight highlight-dark text-left my-6">
+  <figure class="highlight highlight-dark text-left my-6 overflow-auto">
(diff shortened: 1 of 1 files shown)


BinaryMuse:
This closes #2629

HashimotoYT:
> This closes #2629 I think?

Updated with closes

PR #2701, opened 2019-06-26
feat: add error when uses class instead of className for octicons

HashimotoYT:
This PR adds the error when using `class` instead of `className` for giving styles for Octicons.


server.js
```diff
@@ -38,6 +38,10 @@ hbs.registerHelper(lobars)
 hbs.registerAsyncHelper('octicon', async (data, cb) => {
   const { name, className, ariaLabel, width, height } = data.hash
 
+  if (data.hash.class) {
+    return console.log(`Octicons Helper: Use 'className' instead of 'class'.`)
```
malept: Perhaps use the same format as the error below, for consistency:

```suggestion
    return console.error(`ERROR(Octicons Helper): Use 'className' instead of 'class'.`)
```

PR #2847, opened 2019-08-01
feat: add ability to read localized blogs

HashimotoYT:
This PR adds the ability to read blog posts not only in English.

JFYI You can switch to the Japanese language and check how it's works.


package.json
```diff
@@ -8,7 +8,7 @@
   "author": "GitHub",
   "license": "MIT",
   "scripts": {
-    "start": "cross-env NODE_PATH=. NODE_ENV=production node server.js",
+    "start": "cross-env NODE_OPTIONS='--max_old_space_size=4096' NODE_PATH=. NODE_ENV=production node server.js",
```
malept: Is this because it's reading all of the blog post contents into memory?

HashimotoYT: No, we just really load a big JSON file. It reproduced outside of this, in the bump PR it's also crashed.

HashimotoYT:
Simplified to that if we have a non-English language we load from the i18n module, otherwise, load from disk.

HashimotoYT:
ping @zeke 🙇‍♂️

lib/blog.js
```diff
@@ -15,19 +16,20 @@ function parseMarkdown (markdown) {
...
+  static async getAll (language) {
     const files = await fs.readdir(POSTS_PATH)
     const slugs = files.map(file => path.basename(file, '.md'))
-    return slugs.map(BlogPost.get)
+    return slugs.map((slug) => BlogPost.get(slug, language))
   }
 
-  static get (slug) {
-    return new BlogPost(slug)
+  static get (slug, language) {
```
zeke: `BlogPost.get` doesn't seem to add any functionality beyond `new BlogPost`. Is this just here to avoid using the `new` keyword?

HashimotoYT: This is refactored not by me, I'm not able to say the real surface beyond this.

lib/blog.js
```diff
@@ -45,6 +47,24 @@ class BlogPost {
   }
 
   _getRawMarkdown () {
+    // This block looks not very good, but they give confidence,
```
zeke: This file has changed significantly since I was actively involved in the project. What is the purpose of the cache? It seems like this class could be simplified by not using this custom cache, instead using memory as the cache and loading the blog posts at startup/runtime.

HashimotoYT: I think this PR: #2160 can give some feedback around the changes to the blog page handling.

HashimotoYT:
@zeke If solutions used in the refactored version of the blog handler create for you not understandable can you please open an issue where we can discuss the changes. If you do not have any questions about the changes in this PR can we merge them?

PR #2848, opened 2019-08-01
feat: add 'Launch in Fiddle' button to code blocks

deermichel:
Finally the last bit to integrate a `Launch in Fiddle` button to code examples on our webpage.

The button will only be shown if the code block's `data-fiddle-url` html attribute is set (see https://github.com/electron/i18n/pull/769 for more context on how we derive that from the markdown docs).

**To use this feature**, just create a fiddle in the electron repo and add the relative path to the corresponding code block in our `/e/e` docs:
```
```javascript fiddle='{path/to/fiddle/within/e/e}'
const { app } = require('electron')
...
```

cc @felixrieseberg @erickzhao @zeke @HashimotoYT

### DIFF
--- a/data/locale.yml
+++ b/data/locale.yml
@@ -262,6 +262,10 @@ clipboard:
   copy_to_clipboard: Copy to Clipboard
   copied: Copied
 
+fiddle_launch_button:
+  launch: Launch
+  launch_in_fiddle: Launch in Fiddle
+
 or: or
 
 pages:
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ cypress/videos/*
 /precompiled/
 .nyc_output
 coverage
+.DS_Store
--- a/public/styles/hljs/overrides.scss
+++ b/public/styles/hljs/overrides.scss
@@ -1,17 +1,28 @@
 pre {
   overflow: visible !important;
   position: relative !important;
-  .btn-clipboard {
-    outline: 0 !important;
-    position: absolute !important;
-    top: 5px !important;
-    right: 5px !important;
-    padding: .25rem .5rem !important;
+  .btn-group {
+    border: 1px #e5e5e5 solid !important;
+    border-radius: 3px !important;
     font-size: 14px !important;
     font-weight: 500 !important;
-    border: 1px #e5e5e5 solid !important;
+    position: absolute !important;
+    right: 5px !important;
+    top: 5px !important;
+
+    button {
+      border-right: 1px #e5e5e5 solid !important;
+    }
+    button:last-child {
+      border-right: none !important;
+    }
+  }
+
+  .btn-launch, .btn-clipboard {
     background-color: #f8f8f8 !important;
-    border-radius: 3px !important;
+    border: none !important;
+    outline: 0 !important;
+    padding: .25rem .5rem !important;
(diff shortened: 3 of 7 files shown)


data/locale.yml
```diff
@@ -262,6 +262,10 @@ clipboard:
   copy_to_clipboard: Copy to Clipboard
   copied: Copied
 
+launch_button:
```
zeke: Externalized strings! I like it. 🌏

zeke:
🎉 Is there a staging URL for trying this out?

data/locale.yml
DeerMichel: I actually don't know how they'll land in Crowdin - so i just followed the general paradigm 🚀

DeerMichel:
I think https://electron-website-pr-2848.herokuapp.com works - but since there are no docs equipped with a fiddle, you won't notice 👀.

data/locale.yml
HashimotoYT: For more accessibility why, and where this used.

```suggestion
fiddle_launch_button:
```

DeerMichel:
> This doesn't show any helpful information to the user when Fiddle doesn't installed. Maybe redirect to /fiddle?

Yeah, i remember the idea of a launchpad floating around. Unfortunately, i don't think there's currently space on my backlog for it - also i never created sth like that. Maybe @felixrieseberg can give us some detail on how it's done at Slack.

> We can run a specified version of an Electron? Some APIs can be work on 5 but doesn't work at 6.

According to @erickzhao, the version will be derived from the url when you link a file on a specific branch. For a gist, it defaults to the most recent version (but we don't care about that in docs anyway).

DeerMichel:
How about [this](https://github.com/electron/electronjs.org/pull/2848/commits/3f37c4fe89818e555d0d55cb632a7485dbc426e6) approach for now? Simply redirecting the user to the `/fiddle` page when fiddle is not installed. 

Would be nice to add a notice for the user somewhere why this happened tho - any suggestions for a good place :)?

MarshallOfSound:
> How about this approach for now? Simply redirecting the user to the /fiddle page when fiddle is not installed.

For UX reasons we definitely need to send them first to a place that says something like "Attempting to open Electron Fiddle", then when it doesn't appear to work we need to tell them (a) what fiddle is and (b) where to get it.

The page itself should be relatively small and live in the `fiddle.electronjs.org` site.  For example here is the Slack launchpad, a spinner and a few lines of text.

image

DeerMichel:
Oh I like the idea of `Open link in browser` - because that's absolutely possible with gists & repo files. Well, then let's put this launchsite somewhere on my backlog 👀

DeerMichel:
Now using the new http://fiddle.electronjs.org/launch page. This is ready to roll:

jj

deermichel:
Ah the docs on electronjs.org finally were updated again - let me test this :)

deermichel:
Meh, meanwhile got kicked out the electron org and don't have push access anymore - but if you rebase to master, this one works as expected :)

cc @MarshallOfSound

deermichel:
Perfect, ty - ready to be merged by someone with write-access :)

https://electron-website-pr-2848.herokuapp.com/docs/api/screen

PR #3224, opened 2019-11-27
feat: Promote CovalenceConf

felixrieseberg:
This PR adds a little banner for [CovalenceConf](http://www.covalenceconf.com). The banner fades in over 5 seconds.

Here's why I think it deserves a little banner:

- It's the closest thing we have to an official Electron conference
- A bunch of the core contributors will be there
- It's the venue for our official contributor training
- It offers free tickets to those who need them
- It's entirely not-for-profit

image


views/home.html
```diff
@@ -6,6 +6,11 @@
         {{{localized.tagline}}}
       </p>
     </div>
+    <div>
+      <a href="http://www.covalenceconf.com">
```
HashimotoYT: Can be here used `https://`? This url very stands out on main page (start of page). For someone this can be security issue and no one wants to see `Not secure` label aware url.

(This request based on that, what user require for adding app to catalog)

felixrieseberg:
The fade in is done entirely in CSS, so I don’t think there’s a way to make
It faster. I can remove the fade!

As for the url: I’m afraid Covalence currently doesn’t have an https URL :sob:

felixrieseberg:
> As for the url: I’m afraid Covalence currently doesn’t have an https URL 😭

I take that back, we finally got one!

PR #3264, opened 2019-12-11
docs: Add secret announcement

felixrieseberg:
The actual announcement can be found in `#wg-outreach`. I'm seeking approval for this PR so that we can quickly update the page in an hour.


MarshallOfSound:
Seems legit

PR #3424, opened 2020-01-17
remove ugly scrollbar if not needed

noelelias:
overflow auto only adds scrollbars if they are needed otherwise they wont show up overflow scroll always displays them


erickzhao:
Gonna close, seems like this is a duplicate of #2871

PR #3583, opened 2020-02-21
chore: switch to use primer avatar

HashimotoYT:
Just do that what comment says.


HashimotoYT:
Let's thinks it's magic 🧙‍♂️

Issue #3527, opened 2020-02-11
Text breaks on some pages


HashimotoYT:
<img width="994" alt="Screen Shot 2020-02-05 at 10 27 02 PM" src="https://user-images.githubusercontent.com/24681191/74213061-0a545a00-4ca0-11ea-9249-660172a7aaf6.png">

PR #3623, opened 2020-02-29
fix: text breaking on home page

stefankorun:
Fixes text breaking on home page #3527.

**Initial state**
image

**Resulting state**
image


erickzhao:
Closing in favour of #3645, as the `<div>` workaround does not the root problem of CSS specificity.

Thank you for the contribution, though! 🙇

PR #3654, opened 2020-03-05
fix: remove visible code block overflow

erickzhao:
This fixes some CSS that was broken with `.PRIMER-FIX-ME`. Overflow from code blocks generated from markdown now scroll instead of overflowing to the edge of the page.

This is also hopefully help out with SEO, as Google was penalizing us for not being mobile-friendly.

ezgif-6-c676065f83cd

Ancillary changes:
* Refactored selectors to avoid having to use the dreaded `!important`.
* Added more padding to the code blocks such that the "Copy" and "Launch" buttons will never overlap and hide the code below.


public/styles/hljs/overrides.scss
```diff
@@ -1,46 +1,62 @@
-pre {
-  overflow: visible !important;
-  position: relative !important;
+// could remove placeholder selector if we take out
+// PRIMER-REMOVE-ME
+%markdown-pre {
+  overflow: visible;
+  position: relative;
```
malept:
```suggestion
  position: relative;

```

public/styles/hljs/overrides.scss
```diff
@@ -1,46 +1,62 @@
...
+
+.markdown-body pre {
+  @extend %markdown-pre;
+  .PRIMER-REMOVE-ME & {
+    @extend %markdown-pre;
+  }
+}
+
 .hljs {
+  padding: $spacer5 $spacer3;
```
HashimotoYT: The current paddings look for me very big and take the big space between content. I propose to use the new Primer npm padding what's small and beautiful. (The new padding is 24px 12px). Also, it helps with migrating 🙈

```suggestion
  padding: $spacer-5 $spacer-3;
```

erickzhao: Seems like these values are better since they're still big enough to avoid overlap with the top buttons. Thanks!

PR #3696, opened 2020-03-14
chore: Add note about Spectron being unmaintained

justinsilvestre:
Spectron hasn't had any updates for something like a couple years now, save for version bumps and automated minor dependency updates. [This pull request](https://github.com/electron-userland/spectron/pull/345) to update an out-of-date link in the README has been open for more than a year now.

With the current state of the project, I don't think it makes sense to call Spectron "officially supported" without some kind of note that it's not being maintained. At the very least, there should be some kind of warning so that folks aren't encouraged to start new projects using Spectron.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -454,7 +454,7 @@ describe('electronjs.org', () => {
       let $ = await get('/devtron')
       $('.jumbotron-lead .jumbotron-lead-muted').text().should.eq('An Electron DevTools Extension')
       $ = await get('/spectron')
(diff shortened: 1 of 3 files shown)


HashimotoYT:
@erickzhao Announcement banner only works on the main page, but I think it's can be done flexible and make something like this `{{ announcementBanner title="Halp" forceShow=true}}`

ckerr:
Should this be closed as superseded by #5745 ?

PR #3716, opened 2020-03-18
feat: extract handlebars templates to own file

HashimotoYT:
This PR extracts the Handlebars templates and registering to own file for creating some file order.

I create them when working on #3713 for easily managing the other templates (helpers), and don't want to delete them 🙈.

### DIFF
--- a/middleware/hbs-helpers.js
+++ b/middleware/hbs-helpers.js
@@ -1,0 +1,38 @@
+const hbs = require('express-hbs')
+const getOcticons = require('./register-octicons')
+const lobars = require('lobars')
+
+class HandlebarsHelpers {
+  /**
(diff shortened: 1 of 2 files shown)


middleware/hbs-helpers.js
```diff
@@ -0,0 +1,38 @@
+const hbs = require('express-hbs')
+const getOcticons = require('./register-octicons')
+const lobars = require('lobars')
+
+class HandlebarsHelpers {
```
erickzhao: Why does this need to be a class rather than an object?

HashimotoYT: Just convenience during scaling. It could be changed to just to functions.

PR #3852, opened 2020-04-15
chore: add Appzi tracker

erickzhao:
As per discussion in recent wg-ecosystem meetings, this adds the Appzi feedback widget for all our `/docs` page.

Because the widget is only available in English, ~~I've hacked together a gate so that the script only loads on English versions of the website. Not sure if there's a better way to do this (maybe cc @zeke).~~ Thanks to @HashimotoYT, it's now written properly.


erickzhao:
Will gate this to English-only for now as Appzi lacks i18n.

views/partials/head.html
```diff
@@ -52,4 +53,15 @@
...
+    const html = document.getElementsByTagName('html')[0];
+    const head = document.getElementsByTagName('head')[0];
+    const appzi = document.createElement("script");
+    appzi.type = 'text/javascript';
+
+    if (html.lang.startsWith('en')) {
+      appzi.src = 'https://w.appzi.io/bootstrap/bundle.js?token=JW19E';
+      head.appendChild(appzi);
+    }
+  </script>
```
HashimotoYT, line 66:
```suggestion
  {{#eq currentLocale "en-US"}}
  <script async src="https://w.appzi.io/bootstrap/bundle.js?token=JW19E"></script>
  {{/eq}}
```

erickzhao: Much better 😆 

Thanks, didn't know what the syntax was.

PR #3876, opened 2020-04-21
feat: Feature apps on the home page

felixrieseberg:
Our home page is doing a decent job of explaining what Electron is, but it's doing a less good job at telling visitors that Electron is used by applications users _love_. 

Electron is extremely approachable, and as a consequence, the quality of apps out there varies a little bit. We should highlight apps that might make people who'd otherwise cast Electron aside reconsider.

The change looks simple, but I spent a decent amount of time on the design. It handles small and big screens well. 

I've also removed the community apps (they already have a prime spot in the navigation) and the intro video from 2016 on account of it being outdated.

[Also, the usual disclaimer: This isn't a place for apps we love or people we want to honor. This is for apps we hope will make people consider Electron that otherwise wouldn't have. As an example, windows95 was one of GitHub's most popular repos in 2018 and it still wouldn't work here.]


MarshallOfSound:
On thin screen there is no padding between the edge and the heading / paragraph content you added, my first look at this was on my phone so it stood out to me

image

MarshallOfSound:
The logo scaling looks great at 1080p and below but looks strangely large to me on me QHD monitor

image

I think they could be slightly smaller at the upper and lower ends of screen sizes (they're quite big on my phone too).

public/images/featured_apps/teams-icon.svg
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" [
+	<!ENTITY ns_extend "http://ns.adobe.com/Extensibility/1.0/">
```
MarshallOfSound, line 4: Can we minify this SVG (and the others) stripping comments and tags that aren't needed like this.  Running it through something like `svgminify` should help

public/styles/_home.scss
```diff
@@ -24,176 +24,56 @@
...
-  outline: none;
+  justify-content: center;
+  max-width: 90%;
+  margin-left: auto;
+  margin-right: auto;
 }
 
-.listed-app-logo-wrapper {
+a.featured-app-list-item {
+  background-color: #e4e4e4;
```
MarshallOfSound, line 38: Can we do some flex magic on this class to make the boxes have the correct width before the images load.  On my machine when I load the page I can see the images load in asynchronously and it causes this strange stutter effect because these boxes don't have fixed widths but they are based on the images themselves.

Hard to capture a gif but if you scroll down and reload while looking at the images you should see what I mean, kinda a nit but should be easy enough to solve with a `flex: 1` or something

MarshallOfSound:
Implementation looks great, mostly comments around some stylistic stuff and page weight 👍

felixrieseberg:
Alright, here we go! Now that the design is more or less locked down, I made it super-duper-responsive _and_ minimized all the files like crazy. We use `webp` files where possible and fall back to `png`.

The data for all apps weighs less than 1mb, and since you'll never load both `webp` and `png`, it'll be a whole lot less overall. In fact, given that we're cutting out the "old" random apps on the home screen, this PR will reduce the amount of bytes that have to go across the wire.

PR #3932, opened 2020-05-05
chore: switch from standard to eslint + prettier

HashimotoYT:
The same what are we do in other repos. Here used `eslint + prettier` instead of single `prettier` for the ability to use rules.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,10 @@
+module.exports = {
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module'
+  },
+  extends: [
+    'prettier',
+    'plugin:prettier/recommended',
+  ],
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "semi": false,
+  "proseWrap": "always"
+}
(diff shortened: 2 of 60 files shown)


.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
```
erickzhao, line 1: I'm curious, do we have a standard ruleset that we're using across Electron-related projects?

malept: Doesn't seem like it. For example, [`@electron/get`](https://github.com/electron/get/blob/master/.prettierrc.json).

.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module'
+  },
+  extends: [
+    'prettier',
+    'plugin:prettier/recommended',
+  ],
+}
```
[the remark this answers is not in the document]
HashimotoYT: I extracted that to own file since don't love littering `package.json` by these configs.

package.json
```diff
@@ -10,15 +10,15 @@
   "scripts": {
     "start": "cross-env NODE_OPTIONS='--max_old_space_size=4096' NODE_PATH=. NODE_ENV=production node server.js",
     "bootstrap": "node script/bootstrap",
-    "test": "cross-env NODE_PATH=. NODE_ENV=test mocha --timeout 5000 && standard --fix",
+    "test": "cross-env NODE_PATH=. NODE_ENV=test mocha --timeout 5000 && eslint --ext .js --fix",
```
HashimotoYT, line 13:
```suggestion
    "test": "cross-env NODE_PATH=. NODE_ENV=test mocha --timeout 5000 && eslint . --fix",
```

PR #3939, opened 2020-05-08
feat: add ability to use ga events for analytics

HashimotoYT:
This PR adds the ability to use the GA Events to future analyzing the needs of features.

## Usage

For usage, you need to pass `data-sc-category` and `data-sc-name` properties to element. If `data-sc-category` not passed `general` be used as a default category.

Example:

```html
<button data-sc-category="docs" data-sc-name="Something happened">Give some magic</button>
```

---

<details>
<summary>
Original PR Description
</summary>

TL;DL: I just wanna known that is useful before remove that 🤷‍♂️

## Why

So, just why not. Mainly I'm unable to make the GA events work. Also, that creates security since only we can see and maintain that data.

## Usage

For usage, you need to pass `data-sc-category` and `data-sc-name` properties to element. If `data-sc-category` not passed `general` be used as a default category.

After clicking sends the request to the metrics server and save the result to the database. See the [Server](#server) section for more info.

Example:

```html
<button data-sc-category="docs" data-sc-name="Something happened">Give some magic</button>
```

## Server

It's currently nothing is created to accept the request on production. But have something like a small implementation.

### Request

For request used the POST request to `/science` path with the next body:

| Name | Description |
| --- | --- |
| action | The description of the event |
| category | The category of the event |

### Database

The database is used to store the data. The schema of the table:

| Name | Type | Description |
| --- | --- | --- |
| id | int | id of event |
| action | text | the description of the event |
| category | varchar(255) | the category for the event  |
| created_at | Date | timestamp when the event is created |

### Grafana or usage of these data

Currently, I do not create a view of that. But something like `All` and `Per day`?

</details>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,14 +25,15 @@ feel free to propose changes to this document in a pull request.
 - [Context](#context)
 - [Styles](#styles)
 - [Frontend JavaScript](#frontend-javascript)
+- [GA Evens](#google-analytics-events)
 - [Scripts](#scripts)
 - [Environment Variables](#environment-variables)
 
 ## Issues and Pull Requests
 
-* If you're not sure about adding something, [open an issue](https://github.com/electron/electronjs.org/issues/new) to discuss it.
-* Feel free to open a Pull Request early so that a discussion can be had as changes are developed.
-* Include screenshots and animated gifs of your changes whenever possible.
+- If you're not sure about adding something, [open an issue](https://github.com/electron/electronjs.org/issues/new) to discuss it.
+- Feel free to open a Pull Request early so that a discussion can be had as changes are developed.
+- Include screenshots and animated gifs of your changes whenever possible.
 
 ## Commit Messages and Pull Request Titles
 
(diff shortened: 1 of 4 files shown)


HashimotoYT:
I replaced the `fetch` to `ga` function and that's' started working. Previously I used the `onclick` event with `ga` function inside and that doesn't work for me. 

If everybody is fine with GA Events, I can easily replace that without changing using.

CONTRIBUTING.md
```diff
@@ -25,14 +25,15 @@ feel free to propose changes to this document in a pull request.
 - [Context](#context)
 - [Styles](#styles)
 - [Frontend JavaScript](#frontend-javascript)
+- [GA Evens](#google-analytics-events)
```
erickzhao, line 28:
```suggestion
- [Google Analytics Events](#google-analytics-events)
```

CONTRIBUTING.md
```diff
@@ -341,6 +342,22 @@ Like the [Sass middleware](#styles), no compiled JavaScript code is written to
...
 
+## Google Analytics Events
+
+We use the [Google Analytics Events](https://developers.google.com/analytics/devguides/collection/analyticsjs/events)
+for measure purposes of needing and usage of features what are we
+provide on the website.
+
+For usage, you need to provide the `data-sc-category` and `data-sc-name` properties
+to element. If `data-sc-category` is not provided `general` be used as a
+default category for event.
```
erickzhao, line 353:
```suggestion
We use [Google Analytics Events](https://developers.google.com/analytics/devguides/collection/analyticsjs/events)
to measure usage of features that we provide on the website.

To add an event, you need to provide `data-sc-category` and `data-sc-name`
properties to an element in the DOM. If `data-sc-category` is not provided,
`general` will be used as a default category for the event.
```

PR #3968, opened 2020-05-15
build: remove browser-sync

HashimotoYT:
I'm unable to find commit when they are added. Also not fully sure what are we lost in development, but anyway it's not production dependency so everything is ok.


erickzhao:
Is there a particular reason we want to remove it? Seems like it provides something useful for development.

HashimotoYT:
For reducing dependency sizes, and I'm unable to find useful this for yourself in development. How I understand them just sync browser between self.

If that still useful feel free to close that PR.

PR #4050, opened 2020-06-02
Add BLM hompage, show announcement banner on all pages

VerteDinde:
This PR makes two changes, split into separate commits:

1. Updates our announcement banner, so that it is now displayed on all pages. This change adds a property `showAnnouncementBanner` to the context-builder middleware, which can be easily changed in one place. It also updates the documentation to include that property and where to change it. Should we decide we want to keep the announcement banner on all pages long-term, we can leave this commit.
2. Updates our homepage to show a message of solidarity with our community. The commit includes a comment on how to revert back to the homepage when necessary.

Screen Shot 2020-06-02 at 11 05 49 AM

Screen Shot 2020-06-02 at 12 17 06 PM


views/blm.html
```diff
@@ -0,0 +1,42 @@
...
+          <span></span>
+          <p class="h3-mktg text-normal mt-0 mb-4">
+            The Electron community believes that Black Lives Matter. To be silent is to be complicit. 
+            It is our responsibility and our honor to use our platform to amplify the voices of the oppressed and marginalized. 
+          </p>
+          <p class="h3-mktg text-normal mt-0 mb-4">
+            Please donate to <a class="text-white" href="https://blacklivesmatter.com/">Black Lives Matter</a>, 
+            <a class="text-white" href="https://www.naacpldf.org/">the NAACP Legal Defense and Educational
+              Fund</a>, <a class="text-white" href="https://eji.org/">the Equal Justice Initiative</a>,
+              <a class="text-white" href="https://minnesotafreedomfund.org/">the Minnesota Freedom Fund</a>, and
```
felixrieseberg, line 35:
```suggestion
              , and
```

See https://minnesotafreedomfund.org/donate: They are no longer able to accept donations and would like us to promote other organizations (which we're doing here).

VerteDinde: Ah thank you! I'll change that

VerteDinde:
@HashimotoYT Thank you!! I was trying to figure out a cleaner way to do this; will make that change

views/blm.html
sofianguy: Here's a pretty extensive list of ways to donate/help: https://blacklivesmatters.carrd.co/

PR #4058, opened 2020-06-03
fix: Exempt docs from grayscale

Kilian:
This grayscales everything except the docs overview and docs pages, which should help with a11y.


Kilian:
The worst thing that can happen is that something other than docs is also in color again, we can fix that if it comes up.

Kilian:
Thanks @HashimotoYT, Added those too :)

Issue #3261, opened 2019-07-09
youtube_video_url field not displaying video anywhere on app pages


scottmathson:
As [noted in this PR](https://github.com/electron/apps/pull/1119#issuecomment-506823387),  I'm not seeing how the `youtube_video_url` field cited in the [contribution guidelines](https://github.com/scottmathson/apps/blob/master/contributing.md#yml-file-rules) is actually coming into play on-site, on any of these pages.

Dug into other open and closed issues here, alongside searching the source code of pages in production and am not seeing videos displayed anywhere.

/cc @zeke @ckerr

kelvinabella:
> @kelvinabella Go for it!

@HashimotoYT It's my first time contributing to open source. Based from my digging on the issue. We need to add `youtube_video_url` to https://www.electronjs.org/apps/{app-name} if its existing in the yml file. If not so, would appreciate some guidance on where should I start. Thank you!

HashimotoYT:
@kelvinabella For showing the application we use that file: https://github.com/electron/electronjs.org/blob/master/views/apps/show.html. You can start from that point: https://github.com/electron/electronjs.org/blob/b5110284c9d4d490341aa5340685fc5144a39a7c/views/apps/show.html#L171, here we show the screenshots and here should be the youtube video. For checking if the app has youtube video you can use that example:

```hbs
{{#if app.youtube_video_url}}
Show video
{{/if}}
```

kelvinabella:
> @kelvinabella For showing the application we use that file: https://github.com/electron/electronjs.org/blob/master/views/apps/show.html. You can start from that point:
> 
> https://github.com/electron/electronjs.org/blob/b5110284c9d4d490341aa5340685fc5144a39a7c/views/apps/show.html#L171
> 
> , here we show the screenshots and here should be the youtube video. For checking if the app has youtube video you can use that example:
> ```
> {{#if app.youtube_video_url}}
> Show video
> {{/if}}
> ```

@HashimotoYT Hi. Kindly review my changes. Please let me know if I its okay with you. Thanks!

PR #4113, opened 2020-06-17
fix: Adding youtube video to app page

kelvinabella:
Fixing issue [#3261](https://github.com/electron/electronjs.org/issues/3261). Thanks!


kelvinabella:
> Code works for me on Chrome/FF/Safari.
> 
> However, I'm wondering if this should be placed on the left-hand meta bar instead of under the README. The video won't be visible to people unless they scroll all the way down in this current iteration, which seems like a shame.
> 
> cc @HashimotoYT

Hi @erickzhao. 

I'm not sure if I got it correctly. Is this where the video should be placed? Thanks!

image

erickzhao:
@kelvinabella yes, something more like that! I think I would place it on the bottom of the bar rather than the top, but that's what I was getting at. Thanks!

kelvinabella:
@erickzhao Updated. Thanks!

PR #2361, opened 2019-03-28
feat: `@primer/css` migration tracking PR

HashimotoYT:
## Process

All pages described below firstly opens PR targeted into the `master` branch for easily reviewing and managing. In this time this also merges into this branch (`remove-bower-styles`), for final testing without all Bower styles. After that completed this returned back into original PR and after that merged rebased in the current branch. Hard, but it's good for me. 🐱‍💻

## Deploy from this branch

https://electron-website-pr-2361.herokuapp.com/

## Why not bump it by update `@primer/css`? (answer section)

---

### Answers to https://github.com/electron/electronjs.org/pull/2242#issuecomment-478687837

> Do we have a history of Primer breaking in such a way that would require us to completely replace one of its .scss files?

Not sure. We override some Bower styles. Need recheck.

> If not, I'm strongly in favor of just installing @primer/css via npm and using it directly. 

We use `@primer/css` on `/governance` landing page. I'm planning close #757, where wanting deprecating Bower styles and to move to `@primer/css` npm.

> (Also, even if we use @primer/css from npm, we can always copy over styles into the public folder if we really need to.)

I'm _strongly_ don't want to copy styles from `@primer/css` npm package to public folder.

## Breaking changes

:bicyclist: Work In Searching :bicyclist: 

### No more `row`

I do not find a commit where this removed. By it's deprecated. From now it's should be done by flexbox. Time for Flexing.

### `.table`

It's deprecated upstreamely, but it's stays in `electronjs.org`.

### `.container-narrow` and `.container`

:warning: ***WIP*** :warning:

```html
// Old usage
<div class="container-narrow">
	// Content
</div>

// New usage
<div class="container-lg"> // + `p-responsive`
	// Content
</div>
```

### .page-section

```html
// Old usage
<div class="page-section page-section-spaceus">
	// Content
</div>

// New usage
<div class="py-6 py-md-7 py-lg-8" or +1>
	/// Content
</div>
```

## All page tracking progress:

Marked pages merged into `master` and starting rebase into the current branch.

- [x] Homepage (#2346)
- [x] 404
- [x] Apps (#2355)
	- [x] Index
	- [x] Show
- [x] Community (Done in this branch)
- [x] Languages (#2356)
- [x] Donors
- [x] Docs (#2399)
	- [x] Main page
	- [x] Doc page
- [x] Blog (#2388)
	- [x] Index
	- [x] Show
- [ ] Releases (#2423)
- Landings (Done in this branch)
	- [x] Spectron
[...]

PR #4116, opened 2020-06-17
chore: add stylesheet linting

erickzhao:
### Motivation

I was poking around our stylesheets today and noticed that we unfortunately had no consistency in writing style across sheets. Since we already have ESLint for JavaScript linting, this PR adds [stylelint](https://github.com/stylelint/stylelint) for stylesheet linting, as well as a few related changes that help make linting better in general.

Note that none of the soon-to-be-removed `public/bower-components` folder (ref #2361) is linted.

### Changes

* Adds the `stylelint` package to lint stylesheets. See `.stylelintrc.json` for configuration details.
* Adds the `lint-staged` package as a `pre-commit` hook so offending changes can be detected as early as possible.

### Risks

I manually fixed some linting stuff that the `--fix` flag couldn't. Mostly duplicate selector


public/styles/_highlight.scss
```diff
@@ -1,61 +1,61 @@
-.highlight .hll { background-color: #ffffcc }
-.highlight .c { color: #999988; font-style: italic } /* Comment */
-.highlight .err { color: #a61717; background-color: #e3d2d2 } /* Error */
-.highlight .k { color: #000000; font-weight: bold } /* Keyword */
+.highlight .hll { background-color: #ffc; }
```
HashimotoYT, line 1: I'm unable to find that we import or use that file. Also, that looks like we don't own them, let's not lint them.

erickzhao: True, I think I'll just remove the file in a separate PR before merging this one in.

PR #4199, opened 2020-07-07
chore: nav info reorganization

erickzhao:
An attempt at decluttering the top navigation bar in the website so that it's more user-friendly. Groups like items together and removes the "Donors" part of the top nav.

Also aligns the content in the footer with the the elements in the navbar.

cc @VerteDinde


erickzhao:
@HashimotoYT Having "Docs" as the first item of the list (helps reading Left-to-Right, and also with tabbed keyboard navigation) makes the most sense here. I think a common pattern for navbar organization across frameworks that I've seen is that docs (or some other important item that you want your users to visit) is first.

As for Apps vs. Community as the last item, I don't have a strong preference there. I think you're right that it looks visually better the other way around!

erickzhao:
Just pushed an update, hopefully this looks more visually balanaced?

PR #4249, opened 2020-07-20
refactor: remove unused apps icons

HashimotoYT:
These something very old and seems not used. All icons from apps contain from the `electron-apps` module and consumed from them, so I think that would be fine if we remove that duplicate.


HashimotoYT:
@erickzhao #2039 it's just made that more visible and easy to use, previously we use just direct `node_modules/electron-apps` input (well, okay, let's skip that 🙄), that usage starts from #784 and to #2039 where everything starts be more good.

@ckerr I'm really not sure what's that block do (because on that moment I'm idk what's Electron), But if look at this it's reset the screenshot path to `images` folder when unable to find icon inside the `electron-apps` module what's never should be because `electron-apps` give that check on the side, and return all icons. I think in today's state we never should take that state since some modern apps don't have icon inside this repo.

> Some of these are linked to in the old monthly roundup blog entries from 2016.

That icon doesn't work now, and blog posts look not preferable, I can update them to today state or we can 🔥 them

ckerr:
> EDIT: Also that path doesn't more exist from #2039

:1st_place_medal:

PR #4346, opened 2020-08-08
ci: add Node 14 to test matrix

HashimotoYT:
This PR adds Node 14 to the test matrix. Node 14 in around 2-3 months goes to LTS status and we need to create the safely update to the new Node version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,12 +14,16 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x]
+
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js 12
+      - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: ${{ matrix.node-version }}
       - uses: actions/cache@v1
         with:
           path: ~/.npm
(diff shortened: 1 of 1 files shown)


HashimotoYT:
I will release a new version of `cmark-gfm` and `electron-markdown` after some more tests, currently, it's ready for review.

HashimotoYT:
`cmark-gfm` and `electron-markdown` published to npm as `v0.8.0`, all tests in these repositories passed so this should be fine to update.

.github/workflows/test-exploration.yml
```diff
@@ -0,0 +1,62 @@
+name: Test / Exploration
```
malept, line 1: As far as I can tell, it's in a separate file because we don't want to block PRs on Node 14 (or whatever other versions we want to test in the future)? I'd like to think of a clearer name...

.github/workflows/test-exploration.yml
```diff
@@ -0,0 +1,62 @@
...
+      - uses: actions/cache@v1
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+      - run: npm ci
+      - run: npm test
+        env:
+          CI: true
```
malept, line 36: I wonder if we can utilize [composite run steps](https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/) for this file and the existing workflow?

PR #4351, opened 2020-08-10
feat: add support for channel filtering on the RSS feed

MarshallOfSound:
E.g. `/releases.xml?types=stable,beta`

Easy win.  Validated the memoization works so we're not doing significantly more work here


HashimotoYT (approved):
LGTM but maybe in some time this need to be documented somewhere

PR #4357, opened 2020-08-11
fix: remove broken link to the Slack channel

selectiveduplicate:
The current link to the Slack channel is broken; it's now at https://atomio.slack.com/.

### DIFF
--- a/data/locale.yml
+++ b/data/locale.yml
@@ -215,7 +215,7 @@ quick_start:
 
 need_help:
   title: Need Help?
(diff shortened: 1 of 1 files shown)


erickzhao:
cc @electron/wg-outreach regarding our usage of the Atom slack.

bnb:
I would say we should probably remove the link entirely given that we're not really intentionally/actively using the Atom Slack. Could be in this PR or a different one.

PR #4469, opened 2020-09-03
fix: disable the referrerPolicy header

HashimotoYT:
Fixes the bug when open the language navbar on the non-homepage and change the language to another, you be redirected to the home page instead of the same page.


server.js
```diff
@@ -73,6 +73,7 @@ app.use(compression())
 app.use(
   helmet({
     contentSecurityPolicy: false,
+    referrerPolicy: false,
```
erickzhao, line 76: I'm not very familiar with this header. Is there a difference between disabling it with `false` and using `referrerPolicy: { policy: "no-referrer" }`?

erickzhao: Seems like [most browsers use `no-referrer-when-downgrade`](https://web.dev/referrer-best-practices/), and Chrome has `strict-origin-when-cross-origin` as of Chrome 85.

It works for me on Chrome and Safari, so I think this is good to go @HashimotoYT!

erickzhao (approved):

PR #1220, opened 2018-04-05
feat: add sidebar nav to docs pages

zeke:
Resolves #1156 

This PR adds a (localized) sidebar to documentation pages. For tutorial docs, it displays the nested table of contents extracted from [electron/electron/docs/README.md#guides-and-tutorials](https://github.com/electron/electron/blob/master/docs/README.md#guides-and-tutorials).

For API docs, the sidebar displays a list of all APIs.

To Do

- [ ] Make it look better. Currently it's a vanilla unordered list.
- [ ] Fix the left-alignment of the page heading and breadcrumbs.
- [ ] Visually emphasize the currently active link (probably using client-side JavaScript)
- [ ] Display the new nested tutorials nav on the /docs index page too
- [ ] Make it look decent on mobile devices. Maybe the answer is to hide hide the sidebar entirely on mobile.

PR #4475, opened 2020-09-04
feat: make docs page with three columns

vhashimotoo:
This PR changes the docs view component to have the three columns (something like #1220), what's can be used to show the info in the right and left sides. Also that PR started adding the `Did this doc help you?` meta buttons what's being used in the future for more granular tracking of helping (hope?).

How does that look now:

localhost_5000_docs_tutorial_application-debugging

And how that looks with buttons ✌

localhost_5000_docs_tutorial_application-debugging (1)


views/layouts/docs.hbs
```diff
@@ -45,25 +45,58 @@
...
-      </a>
+        <a class="mr-4" href='{{doc.href}}/history'>
+          <span class="octicon octicon-history"></span>
+          {{localized.docs.footer.version_history}}
+        </a>
+      </div>
+    </div>
+
+    <div class="col-sm-2 float-left p-4">
+      {{#if false}}
```
erickzhao, line 78: Are these `#if false` clauses just to disable the inner elements for now?

HashimotoYT: Yes, outer element still be visible because they needed to create the grid.

PR #4592, opened 2020-10-03
refactor: swap raw hex colors to CSS variables

### LINKED ISSUE #4566: Uniformize announcement banner colour codes
We're using raw hex codes here, but should probably be using SCSS variables. Either use existing variables for this banner, or create sensible new ones.

https://github.com/electron/electronjs.org/blob/6772309b3e81a0094581c01a57b37dfaeadece24/public/styles/_home.scss#L145-L156

acrosman:
Fixes #4566 

Mostly uses colors from the existing variable sets and adds one new variable.


public/styles/_home.scss
```diff
@@ -143,8 +143,8 @@ a.featured-app-list-item {
 // Announcement banner------------------------------
 
 .announcement-banner {
-  background: #23252f;
-  border-bottom: 1px solid #191a22;
+  background: $jumbo-bg-color;
```
erickzhao, line 146: I think the announcement banner should probably have more contrast. I'd put `background` to `$jumbo-bg-border` and remove the actual `border-bottom`.

public/styles/_home.scss
```diff
@@ -37,7 +37,7 @@
 }
 
 a.featured-app-list-item {
-  background-color: #e4e4e4;
+  background-color: $main-bg-color;
```
erickzhao, line 40: I think these should stay darker to provide contrast to the (mostly white) screenshots. This change reduces visibility for the screenshots.

image

erickzhao (changes requested):
Thanks for the contribution! A few notes to change here, as the colour codes have changed a bit in this PR.

public/styles/_home.scss
acrosman: Updated per this feedback plus added another color that I missed the first time.

acrosman:
Update to include feedback.  And reversed pulled to ease PR merge.

public/styles/_variables.scss
```diff
@@ -11,12 +11,14 @@ $main-color-strong:   hsl(190, 48%, 24%) !default;
 
 $main-border-color:   hsl(190, 12%, 84%) !default;
 $main-link-color:     hsl(190, 89%, 31%) !default;
+$main-drop-shadow: hsla(0, 0, 28%) !default;
```
erickzhao, line 14:
```suggestion
$main-drop-shadow: hsl(0, 0, 28%) !default;
```

```
[sass]  error: Function hsla is missing argument $alpha.
in /Users/ezhao/Developer/electronjs.org/public/styles/_variables.scss:14:20 
Error: Function hsla is missing argument $alpha.
    at options.error (/Users/ezhao/Developer/electronjs.org/node_modules/node-sass/lib/index.js:291:26)
```

erickzhao (changes requested):

public/styles/_home.scss
erickzhao:
```suggestion
  background: $jumbo-bg-border;
```

erickzhao (changes requested):
@acrosman I think it should be good after this, thanks.

erickzhao (approved):
Thank you!

PR #4604, opened 2020-10-05
feat: add accessibility to homepage link in header

### LINKED ISSUE #4578: Logo in navbar has empty link
There is a logo in the header that links to the home page. The logo is hidden for Assistive Technology like Screen Readers. There is no textual content inside the link. People who use Assistive Technology will not be able to understand where the link leads. This empty link is also not good for SEO.

A possible solution is a visually hidden text inside the link that describes the link purpose.

The corresponding WCAG criterion can be found here: https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html

===

This finding of this issue has been recorded. It is part of my weekly accessibility stream on Twitch: twitch.tv/erikKroes
A recording should be up soon at youtube: https://www.youtube.com/channel/UCwrpGO76k2HyUQspl-rG2eQ/videos

tt-le:
As per WAG criterion: https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context and https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8.

I added an aria-label to the the link describing as "Electron.JS Home Page". 

Now, if a person using assistive technology were to visit this page, the link shall be described as such instead of having no description. 

Fixes #4578


views/partials/header.hbs
```diff
@@ -1,7 +1,7 @@
 <header class="site-header">
   <div class="container">
 
-    <a class="site-header-logo no-underline-hover" href="/">
+    <a class="site-header-logo no-underline-hover" href="/" aria-label="Electron.JS Home Page">
```
erickzhao, line 4:
```suggestion
    <a class="site-header-logo no-underline-hover" href="/" aria-label="Electron Home Page">
```

erickzhao (commented):
LGTM, I think :electron: tends to use `Electron` rather than `Electron.JS`, though.

erickzhao (approved):

Issue #4583, opened 2020-10-01
Installations instructions are marked up as a <figure>


erikkroes:
The installation instructions on the front page are marked up with a figure-element. They are not images but maybe even the most important text on the page for many. This can be improved by putting this content inside an element with different semantics (or no semantics) that don't confuse a user of Assistive Technology.

===

This finding of this issue has been recorded. It is part of my weekly accessibility stream on Twitch: twitch.tv/erikKroes
A recording should be up soon at youtube: https://www.youtube.com/channel/UCwrpGO76k2HyUQspl-rG2eQ/videos

vhashimotoo:
Fixed by #4608

PR #4608, opened 2020-10-06
fix: remove semantic <figure> tags from installation instructions

jasonericdavis:
See issue #4583 for details


erickzhao (approved):
Thank you!

PR #2036, opened 2019-01-31
feat: upgrade Octicons to 8.3.0

HashimotoYT:
This PR updates Octicons to 8.3.0 version, adds a script for automatic updating and removes old Bower Octicons. It uses [icon-font-generator](https://github.com/Workshape/icon-font-generator) for generating icon font.

New icons:

`requested-changes`
`fold-up`
`fold-down`
`eye-closed`
`arrow-both`
`archive`
`kebab-horizontal`
`kebab-vertical`
`project`
`note`
`scrren-full`
`screen-normal`
`grabber`
`plus-small`

PR #4609, opened 2020-10-06
chore: add style linting for helper styles

erickzhao:
Stylesheet linting was originally added in #4116, but the glob in the npm command only ran on files in the root `public/styles` folder.

This PR adds stylesheet linting for all subfolders as well, and fixes existing linting issues.


public/styles/index.scss
```diff
@@ -33,15 +33,13 @@
...
-
 @import "@primer/css/base/normalize.scss";
 @import "@primer/css/base/kbd.scss";
 @import "@primer/css/buttons/index.scss";
 @import "@primer/css/marketing/index.scss";
 @import "@primer/css/tooltips/index.scss";
 @import "@primer/css/utilities/padding.scss";
 @import "@primer/css/pagination/index.scss";
-
 @import "base";
```
vhashimotoo, line 43: We really need space here, this is a divider between primer and our styles.

public/styles/octicons/octicons.scss
```diff
@@ -1,207 +1,209 @@
 @font-face {
```
vhashimotoo, line 1: This file is not ours, can we ignore him?

vhashimotoo (changes requested):

public/styles/octicons/octicons.scss
erickzhao: If it's not ours, does it make sense to move to `/vendor` folder and `.stylelintignore` that folder?

vhashimotoo: What about using the per-file ignore: https://stylelint.io/user-guide/ignore-code#within-files? I not really would move that file to somewhere other location, since not sure what could go wrong.

erickzhao: I'd be okay with doing that for this PR! I think ultimately moving the file to the `/vendor` folder to mark it as `third-party` would make the most sense.

vhashimotoo: He is not fully `third-party`, he just is generated some very old-time ago in #2036, and all part of the code what do the generating removed in #2577, it's one of the things why I wouldn't move this file to `/vendor`. He is is our but not fully ours.

erickzhao: Ooh, thanks for the clarification.

public/styles/index.scss
erickzhao: Added a space here too to clarify

vhashimotoo (approved):

PR #4610, opened 2020-10-06
feat: hero img accessibility

### LINKED ISSUE #4582: Hero image has no alt-text
There is a large SVG at the top of the page in a div named "hero". It does not have an accessible name, but is also not marked up as decorative. From the context I would assume it is decorative. This can be fixed by adding aria-hidden to hide it from Assistive Technology.

More information about the attribute can be found here: https://www.w3.org/TR/wai-aria-1.1/#aria-hidden

===

This finding of this issue has been recorded. It is part of my weekly accessibility stream on Twitch: twitch.tv/erikKroes
A recording should be up soon at youtube: https://www.youtube.com/channel/UCwrpGO76k2HyUQspl-rG2eQ/videos

tt-le:
Added an aria-hidden attribute to the hero div containing a large SVG decorative. This div will be hidden to users using an accessibility software. As per [WAG criterion](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden).

Fixes #4582


erickzhao (approved):
Thank you!

Issue #4584, opened 2020-10-01
Featured apps have superfluous alt text on images


erikkroes:
Every featured app has 2 images (a screenshot and an icon) and a heading text. Both images have alt text that basically indicate what the featured app is. The heading indicates this as well. The images can be marked up with an empty alt-attribute: alt="". This indicates that the images are decorative (because their functionality is already covered by the heading). Right now the accessible name (how the link is pronounced) is very verbose by mentioning the app name thrice.

===

This finding of this issue has been recorded. It is part of my weekly accessibility stream on Twitch: twitch.tv/erikKroes
A recording should be up soon at youtube: https://www.youtube.com/channel/UCwrpGO76k2HyUQspl-rG2eQ/videos

PR #4611, opened 2020-10-07
fix: #4584 featured apps have superflous alt text

jasonericdavis:
Fixes issues #4584 

The output attribute will be **alt** instead of **alt=""** which should be fine according to the specs:
[Section 12.1.2.3 Attributes](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2)


erickzhao (approved):
Thank you!

Issue #4617, opened 2020-10-08
Add "Copy" button to code blocks at Spectron and Devtron pages


vhashimotoo:

stevejoachim:
Hey @vhashimotoo, I added the copy button to both pages. Here's what it looks like:

<img width="883" alt="image" src="https://user-images.githubusercontent.com/19378484/95525034-fed94000-09a0-11eb-994a-9bcaaa3d5f19.png">

PR #4618, opened 2020-10-08
docs: add copy button to spectron and devtron code blocks

stevejoachim:
See #4617 

image

### DIFF
--- a/public/styles/helpers/_code.scss
+++ b/public/styles/helpers/_code.scss
@@ -54,16 +54,17 @@ h4 {
 
   pre {
     padding: $spacer-3;
+    position: relative;
     overflow: visible;
     font-size: 16px;
     line-height: 1.45;
 
     code {
-      display: block;
+      display: inline;
       max-width: initial;
       padding: 0;
       margin: 0;
-      overflow: scroll;
+      overflow: initial;
       line-height: inherit;
       word-wrap: normal;
       background-color: transparent;
--- a/public/styles/helpers/_code.scss
+++ b/public/styles/helpers/_code.scss
@@ -54,16 +54,16 @@ h4 {
 
   pre {
     padding: $spacer-3;
-    overflow: auto;
+    overflow: visible;
     font-size: 16px;
     line-height: 1.45;
 
     code {
-      display: inline;
+      display: block;
       max-width: initial;
       padding: 0;
       margin: 0;
-      overflow: initial;
+      overflow: scroll;
       line-height: inherit;
       word-wrap: normal;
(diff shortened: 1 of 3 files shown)


erickzhao (commented):
screencast 2020-10-09 11-13-57

Getting a weird visual bug on Chrome and FF on macOS (not Safari). 🤔

stevejoachim:
Thanks for catching this! I was able to reproduce this on Firefox. It looks like the Copy button is trying to display a tooltip, but it gets trapped inside the code box and overflows it. I changed some CSS properties for the `pre` and `code` tags to fix this so that the tooltip will display (overflow: visible), but the actual code will still be scrollable (overflow: auto) if the window gets very narrow.

erickzhao (changes requested):
image

Seems like it's gotten visually weirder with the latest push 🤔

stevejoachim:
Thanks for reviewing. That's unfortunate that it's still messed up. What's your test setup? I can't reproduce that visual on Firefox, Safari, or Chrome on MacOS. This is what it's looking like in Firefox for me when hovering on one of the copy buttons:
image

public/styles/helpers/_code.scss
```diff
@@ -54,16 +54,16 @@ h4 {
 
   pre {
     padding: $spacer-3;
-    overflow: auto;
+    overflow: visible;
```
vhashimotoo, line 57:
```suggestion
	position: relative;
    overflow: visible;
```

public/styles/helpers/_code.scss
```diff
@@ -54,16 +54,16 @@ h4 {
...
     font-size: 16px;
     line-height: 1.45;
 
     code {
-      display: inline;
+      display: block;
```
vhashimotoo, line 62:
```suggestion
      display: inline;
```

public/styles/helpers/_code.scss
```diff
@@ -54,16 +54,16 @@ h4 {
...
       max-width: initial;
       padding: 0;
       margin: 0;
-      overflow: initial;
+      overflow: scroll;
```
vhashimotoo, line 66:
```suggestion
      overflow: initial;
```

vhashimotoo (changes requested):
I'm also able to repro the same behavior as @erickzhao on Chrome, Windows 10. The suggestions change behavior to the same as we have at the `/docs` page.

stevejoachim:
@vhashimotoo Thanks for reviewing the code in such detail. I just made the changes you suggested.

PR #4633, opened 2020-10-13
feat: add discord to featured apps

vhashimotoo:
I saw this on Twitter someday about why Discord is not a featured app, so why not to do this, since we already use Discord.


erickzhao (commented):
Agreed here. Seems though that the icon here doesn't fill the entire 70x70 box and thus looks smaller than the other ones.

image

vhashimotoo:
I'm not sure if we can make the icon bigger, this is taken from the official branding guide, and it's the biggest size. But we have also a WhatsApp icon that also looks smaller than others in some cases (the main 150x150, cropped 70x70). For now, I'm not sure what the best way to solve this 🤷‍♂️

PR #4638, opened 2020-10-15
fix: make language picker more accessible

rafalguzniczak:
Fixes https://github.com/electron/electronjs.org/issues/4580
Fixes https://github.com/electron/electronjs.org/issues/4581


erickzhao (approved):
Testing locally on Chrome on macOS:

- [x] `lang-select-button` is a `button` rather than an `a` element.
- [x] `lang-select-button` has an `aria-expanded` attribute to show the menu state.
- [x] `lang-select-button` has a clearer label (done with `aria-label`)
- [x] Keyboard tabbing from closed Languages menu works properly.

vhashimotoo (dismissed):

PR #5003, opened 2021-01-03
fix: remove code block link empty space underlining

kootoopas:
The fact that the `before` and `after` elements occupy some width in conjunction with the underline rule causes surrounding empty space to be underlined too as seen in the attached screenshot, which can in turn cause misunderstandings say in [package names](https://www.electronjs.org/docs/tutorial/quick-start#main-and-renderer-processes) and give one the impression that they are wrapped in underscores, f.e. `_ipcMain_` instead of `ipcMain`.

To clarify this, the horizontal spacing was moved to the `code` element instead.

From:
image

To:
image


erickzhao (approved):
LGTM. The original styles are 5 years old at this point, but this seems simpler and clearer. Thanks for this!

PR #5040, opened 2021-01-11
chore: allowlist i18n languages

erickzhao:
Limits which languages can be viewed on the website. Allowlist currently pending!


cypress/integration/actions.js
```diff
@@ -243,7 +243,7 @@ describe('electronjs.org', () => {
   // For give more information it can contain reproducible steps.
   // It should get value what we should expect for verifying fixing.
   describe('Buggy Tests', () => {
-    it('language bar responsive bug', () => {
+    it.skip('language bar responsive bug', () => {
```
molant, line 246: Is this a flaky test? It's weird that you have to skip it because of the other changes.

erickzhao: I think it seems like this is a test that indicates that a regression bug is still active. I've disabled it because shortening the list of languages makes the `px` values in this test outdated.

molant: Makes sense. Should this be updated before merging?

erickzhao: yeah, I just checked that the bug is still able to repro'd. I'll fix this :)

molant (approved):

PR #5092, opened 2021-01-23
feat: Update command to install brew casks

manosim:
It looks like the way to install a cask with homebrew recently changed.

Running `brew cask install gitify`, would error with:
```
gitify > brew cask install gitify
Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
```

The new way of installing a cask would be by passing --cask as a flag.
```
brew install --cask gitify
```


ckerr:
@manosim please push a new empty commit with a semantic commit message to make the bot happy :slightly_smiling_face:

PR #5097, opened 2021-01-25
fix: memory leak in blog

molant:
The blog was leaking memory because of a race condition with Promises when creating a new `BlogPost`. Because many of the properties relied on the same property, multiple Promises where created to access the same content.

### DIFF
--- a/lib/blog.js
+++ b/lib/blog.js
@@ -45,7 +45,7 @@ async function getAllPosts() {
   const slugs = files.map((file) => path.basename(file, '.md'))
 
   for (const slug of slugs) {
-    blogPosts.set(slug, await getPost(slug))
+    blogPosts.set(slug, await loadPost(slug))
   }
 
   return blogPosts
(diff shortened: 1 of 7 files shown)


molant:
For reference, this is the comparison in terms of memory consumption:
image

lib/blog.js
```diff
@@ -16,115 +22,72 @@ function parseMarkdown(markdown) {
...
-    })
-  }
+  const files = await fs.readdir(POSTS_PATH)
+  const slugs = files.map((file) => path.basename(file, '.md'))
 
-  async frontmatter(key) {
-    const fm = await this._frontmatter()
-    return fm.data[key]
+  for (const slug of slugs) {
+    blogPosts.set(slug, await getPost(slug))
```
ckerr, line 48: I *think* we should be checking the return value for truthiness before adding it here. If the file read fails, this can return null and then we have a map with a valid slug key and and a null value. That will cause a crash later in

```js
    const posts = Array.from(blogPosts.values())
    postsInOrder = posts.sort((a, b) => b.date.localeCompare(a.date))
```

We could filter out null values there, but it's cleaner IMO to not return empty values in the first place.

NB I haven't read the code all the way through. If there's some need to keep the invalid slugs in the map elsewhere in the code, then maybe the `posts` array should just be filtered in that snippet above.

lib/blog.js
```diff
@@ -16,115 +22,72 @@ function parseMarkdown(markdown) {
...
-    return this.frontmatter('title')
-  }
+  let markdown = ''
 
-  date() {
-    return this.frontmatter('date')
+  try {
+    markdown = await fs.readFile(path.join(POSTS_PATH, `${slug}.md`))
+  } catch (e) {
+    blogPosts.set(slug, null)
```
ckerr, line 67:
```suggestion
    blogPosts.delete(slug)
```

ckerr: Outdated

ckerr: Fixed in f5a540e77bff0d02902a0b73126061084405090d

ckerr (approved):
LGTM

PR #5107, opened 2021-01-27
feat: improved API docs tags

erickzhao:
This PR does a couple things:

* Removes some unused SCSS code from 5 years ago.
* Adds labels similar to the platform labels for `Deprecated`, `Experimental`, and `Readonly`.
* Adds link to https://www.electronjs.org/docs/experimental for Experimental tags.


public/styles/_docs.scss
```diff
@@ -114,61 +114,36 @@
   }
 }
 
-.docs-tags {
```
erickzhao, line 117: `.docs-tags`, `.docs-tag`, and `.docs h1 em` were unused relics. I simplified here by modifying the `.platform-label` class and extending it to be able to have different-coloured tags.

bnb (approved):
this is awesome <3

ckerr (approved):
This is a nice look!

PR #5151, opened 2021-02-18
chore: remove appzi

molant:
Removing it as we will be using another solution in the future

### DIFF
--- a/views/partials/head.hbs
+++ b/views/partials/head.hbs
@@ -53,7 +53,7 @@
     ga('create', 'UA-160365006-1', 'auto');
     ga('send', 'pageview');
   </script>
-  {{#eq currentLocale "en-US"}}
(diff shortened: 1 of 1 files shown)


views/partials/head.hbs
```diff
@@ -53,7 +53,7 @@
     ga('create', 'UA-160365006-1', 'auto');
     ga('send', 'pageview');
   </script>
-  {{#eq currentLocale "en-US"}}
-  <script async src="https://w.appzi.io/bootstrap/bundle.js?token=JW19E"></script>
-  {{/eq}}
+  {{!-- {{#eq currentLocale "en-US"}}
```
molant, line 56: I left this intentionally here as we will be using it for the new feedback mechanism

erickzhao (approved):

PR #5153, opened 2021-02-19
HaTS survey

molant:
Current draft for the HaTS survey for the documentation.
How it works at a high level:
- `doc-feedback-cta.js`: It's the client side JS that verifies if the user is a candidate to show the CTA to the survey (running on docs and not having a particular cookie). If the criteria is met, then a request is done to the server to get the survey information. If the server says it's a good candidate, the survey ID is returned and the CTA will be shown (we obviously need a design):

image

- `feedback.js`: The server side component. It's a new route in `/docs/feedback`. It verifies if the survey is open, if the user does not have the cookie (again), and eventually if the user should be shown the CTA (randomly we want to show it to ~8% of the users). If everything checks it returns the survey ID so `doc-feedback-cta.js` can generate the CTA. It also sets the cookie to make sure the user is not shown the CTA again for the next 3 months (unless they delete their cookies).

Pending:

`feedback.js`:
* [x] Get the collector of the latest survey (with caching)
* [x] Implement the logic to check if the survey is opened (with caching)
* [x] Randomly select users
* [x] Cleanup code

`doc-feedback-cta.js`:
* [x] Add a real design to the CTA
* [x] Cleanup code


molant:
Got a lot of things changed in this PR and I think it's actually ready for review. Even though we do not have yet a SM account this should be safe to merge because it will not do anything if the access token is not found.

Video of it working locally:

https://user-images.githubusercontent.com/606594/109346160-1fa61300-7826-11eb-8c1c-bf14c6b2e5c5.mp4

We should also create a few GitHub actions to automatically create surveys, report scores, etc. but I think those can go in a different PR and once we have the account.

@erickzhao the first commit of the PR has information about how everything connects. Happy to jump over hangouts if that helps.

views/partials/feedback.hbs
```diff
@@ -0,0 +1,27 @@
...
+  <style>
+/* By default min height is 300 and position
+     in the center (and can't be tweaked).
+     Also we want only screens with enough
+     space to show the pop up.
+
+     TODO: Might be interesting to animate the
+     entrance instead of just popping up
+  */
+.smcx-widget.smcx-modal {
```
erickzhao, line 14: I think these `!important` rules are unnecessary

views/partials/feedback.hbs
```diff
@@ -0,0 +1,27 @@
...
+  height: 220px !important;
+  bottom: 0 !important;
+  right: 0 !important;
+  top: auto !important;
+  left: auto !important;
+}
+
+.smcx-modal-content {
```
erickzhao, line 22: Could avoid `!important` for this selector by adding more specificity to CSS selectors.

erickzhao (commented):
Aside from the individual questions:

Which style parameters are possible to toggle via the SurveyMonkey settings? In particular, I think that border surrounding the survey is too big 😬

views/partials/feedback.hbs
molant: I believe I had to add it because the CSS from SM was really specific and was added later too so it had higher priority (and there were a few `!important` as well)

molant:
> I think that border surrounding the survey is too big 😬

I might have to override the style as well :(

image

molant:
@erickzhao this should be ready for another review

erickzhao (approved):

PR #5188, opened 2021-03-02
new: survey monkey GitHub action

molant:
**NOTE:** This PR is built on top of #5153. There should be only 1 commit once it is merged and with less changes

This action automatically checks every Monday at 00:00 if a new survey
needs to be created. If so, it creates using the previous one as a model
and closes the old one.

And the video of the action running locally:

https://user-images.githubusercontent.com/606594/109710483-91e26480-7b52-11eb-8241-bfba752aae13.mp4


molant:
@erickzhao FYI, mind 👀?

.github/workflows/electron-survey.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+      - run: npm ci
+      - name: Update survey info
+        env:
+          SM_TOKEN: ${{ secrets.SM_TOKEN }}
+        run: cd .github/actions && node --unhandled-rejections=strict ./create-survey.js
```
molant, line 18: Need to change paths so module resolution works correctly. I didn't do the same as the other action so types are found correctly by VS Code

.github/actions/create-survey.js
```diff
@@ -0,0 +1,58 @@
...
+    const days = Math.floor((survey.dateCreated.getTime() - Date.now()) / oneDay)
+
+    return days >= maxDaysRunning
+}
+
+
+const start = async () => {
+    const lastSurvey = await surveyMonkey.getLatestSurvey()
+
+    if (lastSurvey.id === '-1' || !shouldCreateNewOne(lastSurvey)) {
```
molant, line 37: @erickzhao if you want to test this you will have to replace this condition with:

```js
if (false && (lastSurvey.id === '-1' || !shouldCreateNewOne(lastSurvey))) {
```

erickzhao (approved):

PR #5233, opened 2021-03-16
chore: migrate to docsearch

molant:
WIP on the migration to docsearh:

Current:
Current search results for "main"

docsearch:
Docsearch results for "main"

We get less results with docsearch, but probably more accurate. That said, I think it would be great to have the a bit more granular. Right now everything is under "documentation" and we probably want to add a few more sections (at least API).


molant:
We've done a couple changes to the indexes and there's a preview of this working [in here](https://doodle.com/mm/antonmolleda/electron-documentation) for anyone that wants to provide feedback.

erickzhao (approved):

PR #5251, opened 2021-03-25
fix: more semantic HTML on Releases page

erickzhao:
A couple things here:
* Adding an `<article>` tag around each set of release notes to prevent multiple `<h1>` level tags in a single page context.
* Removing stray `<h3>` tags that were placed ostensibly for visual effect.
* Making the release line menu a `<nav>`.
* Removing `style` attributes on the release line menu, instead putting them in the stylesheet.

### DIFF
--- a/views/releases/index.hbs
+++ b/views/releases/index.hbs
@@ -8,7 +8,7 @@
     </div>
 
     <nav id="release-navbar" class="r-resp-header d-none site-header-nav">
-      <h3>{{localized.releases.show_releases}}</h3>
+      <h1>{{localized.releases.show_releases}}</h1>
       <a class="releases-link-stable {{#eq releasesPage.type 'stable'}}active{{/eq}}" href="/releases/stable">
         <div class="menu-item" style="{{#eq releasesPage.type 'stable'}}border-left: 1px solid #4078C0; background: #fcfcfc{{/eq}}">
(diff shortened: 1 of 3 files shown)


views/releases/index.hbs
```diff
@@ -61,47 +50,47 @@ $ ./node_modules/.bin/electron .</code></pre>
...
+        <article>
+          <h1 class="release-entry" id="{{this.data.version}}">
+            <a class="tag" href="#{{this.data.version}}">Electron {{this.data.version}}</a>
+            <em class="date" data-date="{{this.data.published_at}}" data-format="%B %d, %Y">{{this.data.published_at}}</em>
+            <em class="date ltr-only">(<span data-date="{{this.data.published_at}}"></span>)</em>
+            <a href="{{this.data.html_url}}" class="tag octicon-wrapper"><span class="octicon octicon-mark-github vertical-middle"></span></a>
+          </h1>
+          {{{this.data.body_html}}}
+        </article>
         <hr>
```
molant, line 73: Not sure our use case here applies for [thematic break](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) now that we are using article 🤔

molant (approved):
Lgtm. Have a minor comment but it's more of a question.

views/releases/index.hbs
erickzhao: That's a good point. I can remove that element as well.

PR #5279, opened 2021-03-31
refactor: parse HTML in `/docs` using cheerio

erickzhao:
In preparation for https://github.com/electron/i18n/pull/1786, this PR removes the direct dependency to the `sections` attribute of each document.

Once the new format is in place, we'll need to swap out the concatenation of all section HTML into a single call to the `html` attribute.


routes/docs/index.js
```diff
@@ -1,15 +1,33 @@
 const i18n = require('../../lib/i18n')
+const cheerio = require('cheerio')
 
 module.exports = (req, res) => {
   const docsReadme = i18n.docs[req.language]['/docs/README']
 
+  // Temporary fix to collect all section HTML
+  // Replace with direct HTML when electron-i18n format changes
+  const html = docsReadme.sections.reduce(
```
molant, line 9: Is there a way we can use the final HTML if it's already available or use this fix if it isn't? Maybe by checking if `sections` exists? 
That way we avoid any possible downtime between upgrading the `i18n` package and removing this (or having to do changes to the PR that updates `i18n`).

routes/docs/index.js
```diff
@@ -1,15 +1,38 @@
...
+      // grab all content between headings
+      let content = $(heading)
+        .nextUntil(headingSelector)
+        .map((_, p) => $.html(p))
+        .get()
+        .join()
+      return { html: content }
+    })
+    .get()
+    .slice(2, -1) // slice out unused headings
```
molant, line 33: I was thinking more about explaining what headings are being removed and why (are those sections that we don't need?).

molant (approved):

PR #5326, opened 2021-04-14
chore: close void tags in blog

erickzhao:
Although these closing tags are not required as part of the HTML
specification, certain parsers may fail without the closing tags.


molant:
The contents of the blog live now in https://github.com/electron/electronjs.org-new so this PR is obsolete.

PR #5442, opened 2021-06-08
feat: docs banner

molant:


views/partials/new-docs.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="announcement-banner-contrast">
+  Want to check the beta docs section? Click <a href="/docs/latest/">this link</a>
```
MarshallOfSound, line 2:
```suggestion
  Want to check our <a href="/docs/latest/">beta docs site</a>? We'd love your feedback.
```

Just not a fan of the "Click here" or "use this link" style links 😄

molant:
> could we make it link to the new docs for the current thing you're looking at, rather than to the index page?

@nornagon we thought about it but because there are changes in the folder structure is not obvious to always point to the right place.

At the end we decided it was better to ship sooner, gather feedback/data, and adjust later if needed.

PR #5513, opened 2021-07-03
fix: navbar fixes

AOSXAP:
Fixed too close github button to search bar (on navbar) and also made navbar look better on small displays .

Gallery : https://imgur.com/a/2bQYeT3


erickzhao (changes requested):
Thanks for the PR! Wondering why there's a new package added for a CSS-only fix, though?

AOSXAP:
> Thanks for the PR! Wondering why there's a new package added for a CSS-only fix, though?

sorry , I didn't notice i added that dependency, just made the fix

PR #5548, opened 2021-07-20
feat: add correct link to article in new site

molant:
Thanks to the changes in electronjs.org-new we can now generate a
direct link to the exact article a user is visiting to make the
transition from one to another easier.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Rel https://github.com/electron/electronjs.org-new/commit/cf657d5f30068b14d6d1079c0a2be7c8890ab523

### DIFF
--- a/server.js
+++ b/server.js
@@ -84,6 +84,17 @@ app.use(
     referrerPolicy: false,
   })
 )
+
+// Helper to generate the redirects to the right document in the new docs paths
+hbs.registerHelper('new-docs', (currentPage) => {
+  // This particular page is the root for the docs in the new site
+  if (!currentPage || currentPage.endsWith('tutorial/introduction')) {
+    return '/docs/latest/'
+  } else {
+    return currentPage.replace('docs/', 'docs/latest/')
+  }
+})
+
 if (isProduction) {
   const jsManifest = require(path.join(
     __dirname,
--- a/views/home.hbs
+++ b/views/home.hbs
@@ -97,31 +97,51 @@
 
     <div class="d-sm-flex flex-row text-center text-small mt-3 clearfix">
       <div class="col-xs-12 col-sm-4 col-md-2 offset-md-1 mb-xs-4 mb-md-0 hero-feature">
+        {{#eq currentLocale "en-US"}}
+        <a class="hero-link" href="/docs/latest/api/auto-updater/">
+        {{else}}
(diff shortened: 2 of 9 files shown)


molant:
@nornagon @erickzhao told me you were interested in this feature!

ckerr (approved):
looks good to me.

molant:
@ckerr that's because the new docs live in a separate domain (https://beta.electronjs.org). In the production site we have Fastly that acts as a proxy and any request that goes to /docs/latest hits that domain (internally).
Not sure if I explained my self correctly. Happy to chat 1:1 about this because we will have to do the same for the blog soon-ish!

Also I've added a new "related commit" where I replace the search index with the one used to crawl the new docs so we are pointing more people to the new docs.

server.js
```diff
@@ -84,6 +84,17 @@ app.use(
...
   })
 )
+
+// Helper to generate the redirects to the right document in the new docs paths
+hbs.registerHelper('new-docs', (currentPage) => {
+  // This particular page is the root for the docs in the new site
+  if (!currentPage || currentPage.endsWith('tutorial/introduction')) {
+    return '/docs/latest/'
+  } else {
+    return currentPage.replace('docs/', 'docs/latest/')
```
ckerr, line 94: Are there any docs/latest/ links anywhere that will get adjusted here to docs/latest/latest/ ?

ckerr (commented):
LGTM again

views/home.hbs
```diff
@@ -97,31 +97,51 @@
...
           {{{localized.benefits.native_menus_and_notifications}}}
         </a>
       </div>
       <div class='col-xs-12 col-sm-4 col-md-2 mb-xs-4 mb-md-0 hero-feature'>
-        <a class="hero-link" href="/docs/api/crash-reporter">
+        {{#eq currentLocale "en-US"}}
+        <a class="hero-link" href="/docs/latest/api/crash-reporter">
+        {{else}}
+        <a class="hero-link" href="/docs/latest/api/crash-reporter">
+        {{/eq}}
```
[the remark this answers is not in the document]
molant: Yes, you are right!

Good catch!

server.js
molant: The /docs/latest/ route is handled by the new site and no request will end up in the express server so `currentPage` can never be `docs/latest/*` when reaching this code :)

PR #5650, opened 2021-09-14
chore: deprecate blog routes

molant:
The blog is moving to the new infrastructure. This PR deletes all the
posts, leaves a note, and makes sure the related routes do not render
anything even though they will never be hit in production after Fastly
is updated.


molant:
Before merging this we need to deploy Fastly's configuration!

erickzhao (commented):
looks like we also need to update the tests

molant:
@erickzhao I've done a bit more cleanup: basically deleted more code and changed the feed location.

erickzhao (approved):

PR #5652, opened 2021-09-14
chore: change search index to electronjs

molant:
The current site uses the beta-electronjs index. I've updated the configuration in docsearch to the right entry points so we can go back to the "official" electronjs one.

While the [related PR in docsearch](https://github.com/algolia/docsearch-configs/pull/4586#event-5301918549) has been merged, we have to wait until the index is updated in the next crawl (24h-48h) before this can be merged.


erickzhao (approved):

views/layouts/main.hbs
```diff
@@ -26,13 +26,8 @@
     <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
     <script type="text/javascript"> docsearch({
       apiKey: 'f9fb1d51a99fc479d5979cfa2aae48b9',
```
molant, line 28:
```suggestion
      apiKey: 'c9e8f898b3b32afe40f0a96637e7ea85',
```

PR #5698, opened 2021-10-11
chore: redirect to new docs site for any route

molant:
Redirect to `/latest/` routes now that the localized content is
available.


molant:
We don't want developers to access the old site and we are still seeing a lot of traffic hitting the old doc routes.

PR #5926, opened 2022-04-01
feat: Add an HSTS header for preloading at https://hstspreload.org/

lgarron:
The config format is from https://helmetjs.github.io/

This allows submitting `electronjs.org` to https://hstspreload.org/?domain=electronjs.org

### DIFF
--- a/server.js
+++ b/server.js
@@ -82,6 +82,10 @@ app.use(
   helmet({
     contentSecurityPolicy: false,
     referrerPolicy: false,
+    hsts: {
+      maxAge: 63072000,
+      includeSubdomains: true,
+      preload: true
   })
 )
 


lgarron:
Ahoy! Coming here from the GitHub Product Security Engineering team!

HSTS preloading is an important best practice to make sure that a site is only accessible over HTTPS. This PR would enable preloading in browser to enforce that.

MarshallOfSound (changes requested):
We shouldn't land HSTS or HSTS preload until the migration to the new website is complete.  electron/electronjs.org-new is run behind fastly and will be much easier to configure HSTS at the CDN level rather than on the legacy server which currently only serves the homepage anyway.

